PR #20, opened 2019-07-10
Add Security and Privacy questionnaire and Privacy section

nickburris:
This adds the security and privacy questionnaire for the TAG review, and also adds a section to README on privacy concerns.

### DIFF
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -1,63 +1,63 @@
-2.1. What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
+#### 2.1. What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
 
 The targetText value that will be added to the URL fragment contains a portion of text that is expected to be in the page. This information is either already on the page, or isn't on the page but was expected to be, so shouldn't be a privacy concern.
 
-2.2. Is this specification exposing the minimum amount of information necessary to power the feature?
+#### 2.2. Is this specification exposing the minimum amount of information necessary to power the feature?
 
 The feature itself does not expose information. The URL creator (e.g. a website providing a reference or a user sharing a targetText URL) adds some amount of information to the URL using this feature, but this is information expected to be in the target page.
 
-2.3. How does this specification deal with personal information or personally-identifiable information or information derived thereof?
+#### 2.3. How does this specification deal with personal information or personally-identifiable information or information derived thereof?
 
 It doesn't - The user could theoretically place PII in the targetText, exposing it to the page, but this would imply they expect the information to appear on the page and want to link to it.
 
-2.4. How does this specification deal with sensitive information?
+#### 2.4. How does this specification deal with sensitive information?
 
 It doesn't - See previous question.
 
-2.5. Does this specification introduce new state for an origin that persists across browsing sessions?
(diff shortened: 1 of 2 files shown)


security-privacy-questionnaire.md
```diff
@@ -0,0 +1,63 @@
...
+
+#### 2.11. Does this specification allow an origin some measure of control over a user agent’s native UI?
+
+Similar to an element fragment anchored URL, the origin can cause the UA to change scroll position on navigation. We propose restricting the feature to non-same-doc full navigations, so an origin can't use this to manipulate the scroll position on the current page.
+
+#### 2.12. What temporary identifiers might this this specification create or expose to the web?
+
+None.
+
+#### 2.13. How does this specification distinguish between behavior in first-party and third-party contexts?
```
bokand: I think this is basically referring to cross-origin iframes? In that case, can a cross-origin iframe read the URL of of it's parent page? If so, I think this is something we'd have to address since it would be exposing something about the content of the page that the iframe otherwise couldn't detect. Or am I misunderstanding this section?

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,63 @@
+#### 2.1. What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
+
+The targetText value that will be added to the URL fragment contains a portion of text that is expected to be in the page. This information is either already on the page, or isn't on the page but was expected to be, so shouldn't be a privacy concern.
+
+#### 2.2. Is this specification exposing the minimum amount of information necessary to power the feature?
+
+The feature itself does not expose information. The URL creator (e.g. a website providing a reference or a user sharing a targetText URL) adds some amount of information to the URL using this feature, but this is information expected to be in the target page.
```
bokand: It might be worth mentioning here the restrictions we have in place to prevent exfiltration of page contents.

nickburris: Done.

security-privacy-questionnaire.md
nickburris: Yeah, I didn't fully understand at first but that seems to be what it refers to: https://www.w3.org/TR/security-privacy-questionnaire/#first-third-party

I did some research and also and manually tested this, and cross-origin iframes can only see the parent URL via document.referrer which doesn't include the URL fragment. Updated the answer.

nickburris:
Thanks! Can you merge this? :)

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,66 @@
+#### 2.1. What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
+
+The targetText value that will be added to the URL fragment contains a portion of text that is expected to be in the page. This information is either already on the page, or isn't on the page but was expected to be, so shouldn't be a privacy concern.
+
+#### 2.2. Is this specification exposing the minimum amount of information necessary to power the feature?
+
+The feature itself does not expose information. The URL creator (e.g. a website providing a reference or a user sharing a targetText URL) adds some amount of information to the URL using this feature, but this is information expected to be in the target page. We have restrictions in place to prevent exfiltration of page contents using this feature:
+- Restricted to top-level frames (no iframes)
+- Restricted to non-same-page activations
+- Restricted to pages without an opener (no window.open)
```
bokand: Also user gestures

Issue #15, opened 2019-02-20
Avoid breaking pages that use the URL fragment for routing/state


isiahmeadows:
**Edit:** `s/client instruction/processing instruction/g`

-----

Currently, URIs treat `%%` as invalid. Should we maybe extend accepted URIs to support this use case using that token, as effectively "processing instructions" (like "scroll to text") rather than necessarily seeing it as a fragment? These "processing instructions" would *not* be exposed to users, at least initially, and it wouldn't be included in what's sent to servers.

I'm thinking maybe view it as this:

- `https://example.com#!/route%%q=some%20text` - Search for some text
- `https://example.com#!/route%%q=some%20text%%n=2` - Search for second occurrence of some text
- I could see other potential additions like delaying the search (for JS), selecting or filtering by CSS selector, etc.
- Of course, order doesn't matter. `%%q=some%20text%%n=2` and `%%n=2%%q=some%20text` are equivalent.

Alternatively, you could wrap each "processing instruction" in brackets like `[q=some%20text]` as suggested in #13, but I feel a double percent sign is probably a little easier to explain and use. (I see potential use in both technical and non-technical circles, so accessibility to non-technical people *is* a concern of mine.)

This would resolve and/or address numerous existing issues already filed:

- #2 - This could be as simple as a `%%delay=int_ms` or `%%delay=float_s` instruction.
- #4 - This could be recast as a way to specify Web Annotation [`"target"`](https://www.w3.org/TR/annotation-model/#bodies-and-targets) [selectors](https://www.w3.org/TR/annotation-model/#selectors), but I don't agree we should be bound to their format - it's a bit more verbose than what's necessary for this. We're just selecting crap, and we could include all their functionality with less boilerplate and more flexibility.
- #5 - The long prefix is instead a simple `%%q=`, something even non-technical users could potentially recognize right away due to familiarity with the common convention used by search engines.
- #7 - This doesn't clash. Problem solved. 😉 
- #8 - This could be as simple as a `%%select=div>:nth-child(2)>span.whatever` instruction.
- #9 - This could be as simple as a `%%n=int_n` instruction, as proposed above. We could also include range support here like `1-3` for first three or similar. (Other concerns like "last 4" or "all but first two" need considered.)
- #10 - Unintentionally similar to this, but this is a bit more deliberate and focused where that misses the point of the proposal and goes slightly off-topic.
[...]

PR #21, opened 2019-08-07
Add alternative syntax section to README.md

nickburris:
Add a section on alternative syntax per issue #15.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,14 +42,14 @@ When referencing a specific section of a web page, for example as part of sharin
 
 Allow specifying text to scroll and highlight in the URL fragment:
 
-https://example.com#targetText=prefix-,startText,endText,-suffix
+https://example.com##targetText=prefix-,startText,endText,-suffix
 
 Using this syntax
 
 ```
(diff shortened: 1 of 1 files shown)


yoavweiss:
@nickburris - can you join the WICG and link your GH account with your W3C one to appease the IPR bots?

nickburris:
> @nickburris - can you join the WICG and link your GH account with your W3C one to appease the IPR bots?

Done! (was pending group approval)

@bokand PTAL

bokand:
I think this might a be a bit confusing, a reader could reasonably read this as proposing that either syntax could be accepted. Could we instead update the README to use the new ## syntax?

bokand:
With a caveat that this extension to the URL spec is an open question.

nickburris:
> I think this might a be a bit confusing, a reader could reasonably read this as proposing that either syntax could be accepted. Could we instead update the README to use the new ## syntax?

Done.

bokand:
lgtm, thanks.

PR #26, opened 2019-09-09
Add steps for stripping fragment directive during load

bokand:
First commit formats the bikeshed file to 80 cols

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6,7 +6,9 @@ Shortname: scroll-to-text
 Level: 1
 Editor: Nick Burris, Google https://www.google.com, nburris@chromium.org
 Editor: David Bokan, Google https://www.google.com, bokan@chromium.org
-Abstract: Scroll To Text adds support for specifying a text snippet in the URL fragment. When navigating to a URL with such a fragment, the browser will find the first instance of the text snippet and scroll it into view.
+Abstract: Scroll To Text adds support for specifying a text snippet in the URL
+fragment. When navigating to a URL with such a fragment, the browser will find
+the first instance of the text snippet and scroll it into view.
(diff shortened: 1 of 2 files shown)


bokand:
@nickburris to review

index.bs
```diff
@@ -99,14 +99,40 @@ example" in "here is an example text".
...
+<a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#initialise-the-document-object">
+create and initialize a Document object</a> steps to store and remove the
+fragment directive from the a Document's URL.
+
+Replace steps 7 and 8 of this algorithm with:
+
+7. Let url be null
+8. If request is non-null, then set url to request's current URL.
+9. Otherwise, set url to response's URL.
+10. Find the first instance of the fragment directive delimiter (##) in url's
```
nickburris: Let this be a [position variable](https://infra.spec.whatwg.org/#string-position-variable) and refer to that position in step 11.

I'm not sure if we have to be more specific than "the first instance of the delimiter", see [the infra spec which is as detailed as describing the iteration over the string](https://infra.spec.whatwg.org/#strictly-split). I'm assuming we have a few spots where we need to be more specific algorithmically so we can improve this sort of thing in following iterations.

nickburris:
Looks good to me!

index.bs
bokand: On reflection, I think this part is wrong. url here isn't a string, it's a [URL Record](https://url.spec.whatwg.org/#concept-url) so it's already parsed.

I actually have a followup already in progress that adds the the directive to URL parsing so I'll replace this bit as part of the followup and land this.

PR #28, opened 2019-09-10
Add fragment-directive parsing steps

bokand:
This adds parsing steps for the fragment-directive.

It doesn't correctly account for our current behavior where there is no fragment, e.g.:

https://example.com##targetText=foo

The current steps would require:

https://example.com###targetText=foo

This is foreshadowing a change to the delimiter string which will always require a hash
to put the parser into the fragment state.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -162,6 +162,18 @@ the fragment is the string "test" and the fragment-directive is the string
 "targetText=foo".
 </div>
 
+### Serializing the fragment directive
+
+Amend the <a href="https://url.spec.whatwg.org/#url-serializing">URL serializer
+</a> steps by inserting a step after step 7:
+
+<em>
+8. If the exclude fragment flag is unset and url's fragment-directive is
+non-null:
+    1. If url's fragment is null, append U+0023 (#) to output.
+    2. Append "##", followed by url's fragment-directive, to output.
+</em>
+
 ### Processing the fragment directive
 
 To the definition of
--- a/index.bs
+++ b/index.bs
@@ -167,12 +167,10 @@ the fragment is the string "test" and the fragment-directive is the string
 Amend the <a href="https://url.spec.whatwg.org/#url-serializing">URL serializer
 </a> steps by inserting a step after step 7:
 
-<em>
-8. If the exclude fragment flag is unset and url's fragment-directive is
+8. If the <em>exclude fragment flag</em> is unset and <em>url's fragment-directive</em> is
 non-null:
-    1. If url's fragment is null, append U+0023 (#) to output.
-    2. Append "##", followed by url's fragment-directive, to output.
-</em>
+    1. If <em>url's fragment</em> is null, append U+0023 (#) to <em>output</em>.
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -111,28 +111,94 @@ author script so that future UA instructions can be added without fear
...
+<em>
+A URL's fragment-directive is either null or an ASCII string holding data used
+by the UA to process the resource. It is initially null
+</em>
+
+Amend the <a href="https://url.spec.whatwg.org/#concept-basic-url-parser">
+basic URL parser</a> steps to parse fragment directives in a URL:
+
+  - In step 11 of this algorithm, amend the <em>fragment state</em> case:
+    - In the inner switch on <em>c</em>, add a step after step 2:
```
nickburris: - In the inner switch on <em>c</em>, **in the Otherwise case,** add a step after step 2:

index.bs
```diff
@@ -111,28 +111,94 @@ author script so that future UA instructions can be added without fear
...
+        - If <em>c</em> is U+0023 (#) and <em>remaining</em> begins with U+0023
+          (#), set state to <em>fragment-directive state</em>. Increment <em>c</em>
+          by the length of the fragment-directive delimiter minus 1. <div class="note">
```
nickburris: Define fragment-directive delimiter somewhere?

index.bs
```diff
@@ -111,28 +111,94 @@ author script so that future UA instructions can be added without fear
...
+    - Switching on <em>c</em>:
+        - The EOF code point: Do nothing
+        - U+0000 NULL: Validation error
+        - Otherwise:
+          1. If <em>c</em> is not a URL code point and not U+0025 (%), validation
+             error.
+          2. If <em>c</em> is U+0025 (%) and <em>remaining</em> does not start with
+             two ASCII hex digits, validation error.
+          3. UTF-8 percent encode <em>c</em> using the fragment percent-encode set
+             and append the result to <em>url’s fragment-directive</em>.
```
nickburris: We don't want to percent encode '-', or ',' (actual usage of these characters must be percent encoded already so we can interpret them as fragment directive syntax).

index.bs
bokand: Done.

index.bs
bokand: Done

bokand: Also made a note that this is likely to change.

index.bs
bokand: I think we're fine here? https://url.spec.whatwg.org/#fragment-percent-encode-set doesn't include '-' or ','.

bokand:
Sorry, I messed up (somewhat new to pull-requests and inappropriately used the master branch) so this PR now includes a followup I had. Not sure if you saw that or not but I've also included the steps for serialization. I'm not sure if I can cleanly fix it in this PR so could you just take a second look at that part?

I'll try to split things up in future PRs better.

nickburris:
Yep, I saw those changes as well, LGTM!

PR #29, opened 2019-09-10
Improve target text matching algorithm

nickburris:
Improved the target text matching algorithm, which aligns with the structure of the Chromium implementation. Also fixed all bikeshed formatting errors and warnings.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,0 +1,30 @@
+<pre class='metadata'>
+Status: CG-DRAFT
+Title: Scroll To Text Fragment
+ED: wicg.github.io/ScrollToTextFragment/draftspec.html
+Shortname: scroll-to-text
+Level: 1
+Editor: Nick Burris, Google https://www.google.com, nburris@chromium.org
+Editor: David Bokan, Google https://www.google.com, bokan@chromium.org
+Abstract: Scroll To Text adds support for specifying a text snippet in the URL fragment. When navigating to a URL with such a fragment, the browser will find the first instance of the text snippet and scroll it into view.
+Group: wicg
+Repository: wicg/ScrollToTextFragment
+</pre>
+
+# Introduction # {#introduction}
+TODO
+
+## Use cases ## {#use-cases}
+
+### Web text references ### {#web-text-references}
+The core use case for scroll to text is to allow URLs to serve as an exact text reference across the web. For example, Wikipedia references could link to the exact text they are quoting from a page. Similarly, search engines can serve URLs that direct the user to the answer they are looking for in the page rather than linking to the top of the page.
+
+### User sharing ### {#user-sharing}
+TODO
+
+# Description # {#description}
+TODO - description of amendment to [[html#scroll-to-fragid]]
+
+<div class="example">
+https://github.com/WICG/ScrollToTextFragment#targetText=future%20extension.-,we%20allow%20highlighting,other%20targetTexts.,-e.g.
+</div>
--- a/index.bs
+++ b/index.bs
@@ -7,8 +7,8 @@ Level: 1
 Editor: Nick Burris, Google https://www.google.com, nburris@chromium.org
 Editor: David Bokan, Google https://www.google.com, bokan@chromium.org
 Abstract: Scroll To Text adds support for specifying a text snippet in the URL
-fragment. When navigating to a URL with such a fragment, the browser will find
-the first instance of the text snippet and scroll it into view.
+    fragment. When navigating to a URL with such a fragment, the browser will
+    find the first instance of the text snippet and scroll it into view.
 Group: wicg
 Repository: wicg/ScrollToTextFragment
 </pre>
@@ -111,7 +111,7 @@ author script so that future UA instructions can be added without fear
 introducing breaking changes to existing content. Potential examples could be:
 translation-hints or enabling accessibility features.
 
-### Processing the fragment directive
+### Processing the fragment directive ### {#processing-the-fragment-directive}
 
 We amend the
 <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#initialise-the-document-object">
@@ -147,9 +147,10 @@ href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-indicated
 indicated part of the document</a>.
 
 1. Let <em>fragment directive</em> be the document URL's
-<a href="#fragment-directive">fragment directive</a>. 
+    <a href="#fragment-directive">fragment directive</a>. 
 2. If [[#find-a-target-text]] with <em>fragment directive</em> returns
-non-null, then the return value is the indicated part of the document; return.
+    non-null, then the return value is the indicated part of the document;
+    return.
 
 ### Find a target text ### {#find-a-target-text}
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -111,7 +111,7 @@ author script so that future UA instructions can be added without fear
 introducing breaking changes to existing content. Potential examples could be:
 translation-hints or enabling accessibility features.
 
-### Parsing the fragment directive
+### Parsing the fragment directive ### {#parsing-the-fragment-directive}
```
bokand: I cargo-culted this when I was working on it but it kept showing up in the output - what does this do and how do you prevent it from showing up in the rendered html?

index.bs
```diff
@@ -219,71 +219,124 @@ href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-indicated
...
+    page.
+    <div class="note">
+    This algorithm returns a <em>text range</em>, defined as a text position and
+    text length that specifies a text fragment in the page text. It is up to the
+    implementation how this text range within <em>page text</em> will map to the
+    indicated part of the document.
+    </div>
+6. Let <em>position</em> be a position variable for <em>page text</em>,
+    pointing at the start of <em>page text</em>.
+7. If <em>tokens</em> has size 1, then:
```
bokand: Isn't this subsumed by steps 8 and on? i.e. I think the exact match case is handled by the algorithm that handles the other optional cases, no?

index.bs
```diff
@@ -219,71 +219,124 @@ href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-indicated
...
+        1. Advance <em>search position</em> to the end of the next instance of
+            <em>prefix</em> in <em>text</em>
+        2. If <em>search position</em> points past the end of <em>text</em>,
+            then break.
+        3. If the text following <em>search position</em> does not equal
+            <em>query</em>, then continue.
+    2. Let <em>potential match position</em> be a position variable equal to
+        <em>search position</em>.
+    3. Advance <em>search position</em> to the end of the next instance of
+        <em>query</em> in <em>text</em>.
```
bokand: Is this line assuming that _search position_ points to the beginning of query and thus this moves us to the end of it? It so, an assert or note here might be helpful.

bokand: Also, if so, doesn't that mean the line below will never happen because in 1.1.3 we would have continued if the following text isn't query.

index.bs
nickburris: It explicitly defines the element ID of the header, bikeshed generates one anyway based on the text but gives a warning that it should be explicitly defined. I'm not sure why it would show up in the rendered html, maybe we're using different versions of bikeshed? I'm just using the web api, `curl https://api.csswg.org/bikeshed/ -F file=@index.bs -F force=1 > index.html`, or `curl https://api.csswg.org/bikeshed/ -F file=@index.bs -F output=err` to get warnings and errors.

index.bs
```diff
@@ -219,71 +219,124 @@ href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-indicated
...
 4. Let <em>tokens</em> be a list of strings that is the result of splitting the
     string <em>raw target text</em> on commas.
-5. If <em>tokens</em> has size 1, then:
+5. Let <em>page text</em> be a string that is all of the visible text on the
+    page.
+    <div class="note">
+    This algorithm returns a <em>text range</em>, defined as a text position and
+    text length that specifies a text fragment in the page text. It is up to the
+    implementation how this text range within <em>page text</em> will map to the
+    indicated part of the document.
```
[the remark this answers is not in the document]
tilgovi: Definitely needs to be specified, I think.

Do we want the text to be affected by things like `text-transform` or white-space collapsing?

I've punted on all this in the past when doing document highlighting and text search because it was so much cheaper to get `body.textContent` than to use a `Selection.prototype.toString()` to get something more accurately resembling what the user sees.

index.bs
bokand: I'm using the web API too...maybe I just had a typo?

index.bs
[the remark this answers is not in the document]
bokand: > Do we want the text to be affected by things like `text-transform` or white-space collapsing?

We've made the text search case insensitive so that we avoid issues with `text-transform`. I think this might be a difference with TextQuoteSelector?

I've noticed WebAnnotations refers to "normalizing" text but can't find that defined anywhere. Do you have a pointer? Also, I think I've seen something about how WebAnnotations does white-space collapsing but can't find it now, would you have a pointer?

> I've punted on all this in the past when doing document highlighting and text search because it was so much cheaper to get body.textContent than to use a Selection.prototype.toString() to get something more accurately resembling what the user sees.

That's an interesting approach. Do you know roughly how often it happens that the text presentation differs significantly from the markup?

index.bs
```diff
@@ -219,71 +219,124 @@ href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-indicated
...
     </div>
 3. If <em>raw target text</em> is the empty string, return null.
 4. Let <em>tokens</em> be a list of strings that is the result of splitting the
     string <em>raw target text</em> on commas.
-5. If <em>tokens</em> has size 1, then:
+5. Let <em>page text</em> be a string that is all of the visible text on the
+    page.
+    <div class="note">
+    This algorithm returns a <em>text range</em>, defined as a text position and
+    text length that specifies a text fragment in the page text. It is up to the
```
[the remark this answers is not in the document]
nickburris: Done (text range to be consistent with the rest of the doc)

index.bs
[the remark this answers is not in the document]
nickburris: I added a TODO here. I'm not sure there's precedence for this, window.find isn't specified unfortunately.

index.bs
nickburris: That's right, step 7 is exactly the same as step 16, removed this.

index.bs
nickburris: That's the case if a prefix is specified, but if there is no prefix then this step just advances to the next instance of query. I added a note describing the two cases.

bokand: ah, right. Thanks!

bokand: Thanks, I missed that. But, suppose there's no prefix or suffix:

2. will set `potential match position` to `search position` which will point to some potentially random part of the text
3. `search position` now points to the end of query
5. Will now return `potential match position` which doesn't point to query, it just points to the value of `search position` at the start of the algorithm.

nickburris: That's right, thanks! Fixed.

index.bs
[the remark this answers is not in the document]
tilgovi: I broke spec in tilgovi/dom-anchor-text-position and tilgovi/dom-anchor-text-quote because I didn't see an easy way to get the whitespace-normalized text for the full page.

The text presentation differs significantly with regard to whitespace, because often elements have leading or trailing whitespace on their text content that gets collapsed, or leading whitespace on each line if the text content is authored with line breaks.

Related, there is also the CSS `white-space` property.

Seeing all this, I just punted entirely for my anchoring libraries and worked on the raw `textContent` of the body.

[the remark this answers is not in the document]
tilgovi: I guess that `innerText` is specified these days and well supported, so maybe that's the ticket I was missing to avoid using `Selection` to extract the text.

index.bs
```diff
@@ -317,18 +317,18 @@ This algorithm has input <em>text, search position, prefix, query,</em> and
             then break.
```
bokand: When we say 

> Advance search position to the end of the next instance of prefix in text

Does this mean we point to the last char of prefix or the char after it? If it's the char after, step 2. below is now wrong in the `prefix != null` case because search position will point to an instance of query. So I expect

> Advance search position to the end of the **next** instance of query in text

Would skip the query instance we're currently at the beginning of and find the next one, then go to its end. I expect this is the case based on 

> Let potential match position be a position variable equal to search position minus the length of query.

This would only work if "end of instance" means one past the last char.

index.bs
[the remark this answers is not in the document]
tilgovi: I think I understand you correctly.

Using `innerText` makes it hard to create a `Range` because the offsets of a `Range` are measured against the text content rather than the inner text.

This may be another reason why I chose not to use `innerText`.

index.bs
nickburris: By "advance to the end of the next instance of prefix" I do mean point to the char after it, but I also think "advance to the end of the next instance of query" means advance to the end of query if we're currently pointing to the beginning of query. Maybe "advance _search position_ to the end of the first instance of _query_ in _text_ starting from _search position_" is more clear?

PR #41, opened 2019-10-04
Specify word boundary matching

nickburris:
One of the [security restrictions](https://github.com/WICG/ScrollToTextFragment#security) is word boundary matching. This PR specifies how the word boundary matching is incorporated into the matching algorithm.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -14,7 +14,6 @@ Repository: wicg/ScrollToTextFragment
 </pre>
 
 # Introduction # {#introduction}
-TODO
 
 ## Use cases ## {#use-cases}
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -367,3 +379,33 @@ a copy, i.e. any modifications are performed on the caller's instance of
...
+    4. Using locale <em>locale</em>, let <em>right bound</em> be the first word
+        boundary in <em>text</em> after <em>range</em>.
+        <div class="note">A word boundary is as defined in the 
+        <a href="http://www.unicode.org/reports/tr29/#Word_Boundaries">Unicode
+        text segmentation annex</a>. The
+        <a href="http://www.unicode.org/reports/tr29/#Default_Word_Boundaries">
+        Default Word Boundary Specification</a> defines a default set of what
+        constitutes a word boundary, but as the specification mentions, a
+        language-specific word boundary library should be used where possible
+        based on the <em>locale</em>, such as the ICU library.
```
bokand: I think we should add a note here that 1) word boundaries are part of the security mitigations and 2) for dictionary based approaches in languages with a small number of characters (<100?) the dictionary must not contain a large proportion of the individual characters (<20%?).

Might also be worth adding a note linking to https://github.com/tc39/proposal-intl-segmenter and saying this might eventually be specified elsewhere in the platform.

bokand:
I'll merge this, please add the requested note as a followup

PR #43, opened 2019-10-08
Change wording around word boundary spec

bokand:
Reworded the section around word boundary computation and its details around languages without a word separating character (space).


bokand:
On re-reading, I decided to reword some of the new section, WDYT? In particular, I think referring to a specific library in the spec is probably a bit too specific so I've removed the reference to ICU.

PR #48, opened 2019-10-10
Add note to spec about directive appearing in fragment

bokand:
Add a note and example to clarify that the fragment-directive must always appear in the fragment
portion of the URL. i.e. To generate a text directive, a hash must be added to the URL if it doesn't contain an existing fragment.


bokand:
Closing in favor of branch-based PRs

PR #49, opened 2019-10-10
[Spec] Add recommendations for generating URLs

### LINKED ISSUE #45: Add recommendation in spec for UAs that generate :~:text= links
We should add some guidelines/recommendations to the spec for best-practices around generating `:~:text=` links. We shouldn't make these MUST statements since UAs should be free to experiment and iterate on how they generate links, so long as they conform to the specified syntax; however, having some guide-rails will help ensure a basic level of user friendliness.

Some examples:

- Stripping fragment-ids from the URL before appending a `:~:text=` directive. This could be done optionally, if the element fragment has been scrolled far enough off-screen
- Guidelines around when to prefer including the entire snippet vs. range-based
- How/When to use context terms

bokand:
Added some initial non-normative text about best-practices in generating text fragment directives.

Fixes #45

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -542,10 +542,10 @@ as a range-based match.
 
 Context terms allow the text fragment directive to disambiguate text snippets
 on a page. However, their use can make the URL more brittle in some cases.
-Often, the desired string will at an element boundary. The context will
-therefore exist in an adjacent element. Changes to the page structure could
-invalidate the text fragment directive since the context and match text may no
-longer appear to be adjacent.
+Often, the desired string will start or end at an element boundary. The context
+will therefore exist in an adjacent element. Changes to the page structure
+could invalidate the text fragment directive since the context and match text
+may no longer appear to be adjacent.
 
 <div class='example'>
   Suppose we wish to craft a URL for the following text:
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -441,4 +441,157 @@ Location Interface</a> to include a fragmentDirective property:
...
+
+<div class='note'>
+  TODO:  Can we determine the above limit in some more objective way?
+</div>
+
+## Use Context Only When Necessary
+
+Context terms allow the text fragment directive to disambiguate text snippets
+on a page. However, their use can make the URL more brittle in some cases.
+Often, the desired string will at an element boundary. The context will
```
nickburris: "will **start or end** at an element boundary"

index.bs
```diff
@@ -441,4 +441,157 @@ Location Interface</a> to include a fragmentDirective property:
...
+
+<div class='example'>
+  Suppose we wish to craft a URL for the following text:
+
+  <pre>
+        &lt;div class="section"&gt;HEADER&lt;/div&gt;
+        &lt;div class="content"&gt;Text to quote&lt;/div&gt;
+  </pre>
+
+  We could craft a the text fragment directive as follows:
```
nickburris: Remove "a"

index.bs
bokand: Fixed

index.bs
bokand: Fixed

Issue #36, opened 2019-09-25
Editorial: increase the number of cross-links


domenic:
It'd make the spec easier to read if there were a liberal sprinkling of cross links to relevant concepts. For example:

- Lots of Infra concepts are used. Although it's less important to link to basic things like "string starts with" or "remove from a list" or similar, some of the stuff around position variables would be helpful.
- "Find an exact match with context" is italicized, not linked
- "Each document has an associated fragment directive" should have a `<dfn>`, and further references should link to it. Similarly for URL record.
- References to "c" in the URL parser should go to https://url.spec.whatwg.org/#c

One heuristic is that you seem to have italicized a lot of things that should be links. Maybe do an audit of all your italicizations.

PR #56, opened 2019-10-17
Add formal grammar for text directives

### LINKED ISSUE #32: Add formal conformance/writing requirements for web developers
You have the most important part, the URL record and the parsing modifications. This is all browsers need. But it's also important to update the definition of a URL for web developers, in https://url.spec.whatwg.org/#url-writing

Maybe nothing changes in that section actually, since the same set of strings is valid before and after?

However, it may be good to have a section describing, in a similar rigorous fashion, the validity of the targetText directive in particular. Some of this is done informally in https://wicg.github.io/ScrollToTextFragment/#syntax. Such a section would probably end up in HTML (or wherever the actual targetText directive is defined), instead of being part of the general URL infrastructure.

nickburris:
Fixes #32 by adding a formal grammar for text directives. Also modified a couple of terms to be proper `<dfn>`s ("fragment directive" and "fragment directive delimiter"), note there are many other terms that should be `<dfn>`s, I'll clean these up in a separate PR addressing #36.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -217,12 +217,25 @@ Replace steps 7 and 8 of this algorithm with:
 11. Set <em>url's fragment-directive</em> to null.
 12. Set the <em>document's url</em> to be <em>url</em>.
 
-### Text directive grammar ### {#text-directive-grammer}
-A <dfn>valid text directive</dfn> is a sequence of characters that appears in
-the [=fragment directive=] that matches the production:
+### Fragment directive grammar ### {#fragment-directive-grammar}
+A <dfn>valid fragment directive</dfn> is a sequence of characters that appears
+in the [=fragment directive=] that matches the production:
+<dl>
+<code><dt><dfn>FragmentDirective</dfn> ::=</dt> <dd>[=TextDirective=] ("&" [=TextDirective=])*</dd></code>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -214,6 +217,26 @@ Replace steps 7 and 8 of this algorithm with:
 11. Set <em>url's fragment-directive</em> to null.
 12. Set the <em>document's url</em> to be <em>url</em>.
 
+### Text directive grammar ### {#text-directive-grammer}
+A <dfn>valid text directive</dfn> is a sequence of characters that appears in
+the [=fragment directive=] that matches the production:
+
+<dl>
+<code><dt><dfn>TextDirective</dfn> ::=</dt> <dd>"text=" [=TextDirectiveParameters=] ("&text=" [=TextDirectiveParameters=])*</dd></code>
```
bokand: I think the (&text=...)* part should probably appear separately as the rules for a valid _fragment_ directive.

That is, assuming future directive types, we'd want to allow:

```
:~:text=foo&widget=bar&text=baz
```

Which I think this definition precludes.

nickburris:
Agreed - opened #57 to track

PR #58, opened 2019-10-21
Increase pointer, not c

### LINKED ISSUE #33: "Increment c"?
> Increment c by the length of the fragment-directive delimiter minus 1.

Nothing in the URL spec increments c. It's unclear what this would mean; c is a character, so is incrementing it by 1 suppose to give the character that's 1 larger in the Unicode code point space?

I suspect you want to "increase pointer by..." which is a phrase that is used elsewhere.

nickburris:
To advance `c` we increase `pointer` as the URL spec does. By definition, `c` is the character currently at `pointer`. Fixes #33.


bokand:
lgtm

PR #65, opened 2019-10-25
Move fragment directive parsing to Document initialization

### LINKED ISSUE #60: Move fragment-directive parsing into HTML
Based on the discussion in whatwg/url/issues/445, I think we should move the sections of the spec dealing with parsing the fragment directive from the URL into the HTML spec only, so that it's only affected as part of HTML document loading.

nickburris:
This patch moves the fragment directive parsing steps out of the URL parser and into the document initialization steps.

One caveat is that the URL serialization can no longer refer to the URL's stored fragment directive. My solution is to refer to the "associated Document"'s fragment directive, but I'm open to other ideas here.

Fixes #60

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -120,7 +120,7 @@ translation-hints or enabling accessibility features.
 ### Parsing the fragment directive ### {#parsing-the-fragment-directive}
 
 To the definition of
-<a href="https://dom.spec.whatwg.org/#concept-document-type">Document</a>, add:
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -192,41 +195,14 @@ the fragment is the string "test" and the [=fragment directive=] is the string
 Amend the <a href="https://url.spec.whatwg.org/#url-serializing">URL serializer
```
bokand: I think we can remove this (serialization) section entirely now

PR #67, opened 2019-11-04
Specify how a Range is the indicated part of the document

### LINKED ISSUE #66: Clarify how scroll to fragment is performed
https://wicg.github.io/ScrollToTextFragment/#navigating-to-text-fragment defers to HTML5 in order to perform the scroll to the fragment.

https://html.spec.whatwg.org/multipage/browsing-the-web.html#scroll-to-the-fragment-identifier assumes that the indicated part of the document is an Element and then defers to CSSOMView.
However, https://wicg.github.io/ScrollToTextFragment/#find-a-target-text returns a Range.

Maybe what is wanted instead is to rely on window.find() but this is not standardized. See https://github.com/whatwg/html/issues/3539

nickburris:
This patch specifies how a Range is used as the indicated part of the document. It refactors the "scroll rect into view" part of the CSSOM's scroll element into view algorithm so that it can be reused in an algorithm scrolling a Range into view. _The indicated part of the document_ can now be specified as a Range in addition to an Element, where the element becomes the Document's target element and the Range is used for the scroll into view position.

Fixes #66

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -320,8 +320,8 @@ The scroll to text specification proposes an amendment to
 [[html#scroll-to-fragid]]. In summary, if a [=text fragment directive=] is
 present and a match is found in the page, the text fragment takes precedent over
 the element fragment as the indicated part of the document. We amend the
-indicated part of the document to possibly include a [[DOM#range|Range]] that is
-scrolled into view instead of the containing element.
+indicated part of the document to optionally include a [[DOM#range|Range]] that
+is scrolled into view instead of the containing element.
 </div>
 
 Replace step 3.1 of the [[HTML#scroll-to-the-fragment-identifier|scroll to the
@@ -337,8 +337,8 @@ fragment]] algorithm with the following:
 3. Otherwise:
     3. If <em>range</em> is non-null:
         1. [=scroll a Range into view|Scroll range into view=], with
-            <em>behavior</em> set to "auto", <em>block</em> set to "start", and
-            <em>inline</em> set to "nearest".
+            containingElement <em>target</em>, <em>behavior</em> set to "auto",
+            <em>block</em> set to "start", and <em>inline</em> set to "nearest".
     4. Otherwise:
         1. [[cssom-view#scroll-an-element-into-view|Scroll target into view]],
             with <em>behavior</em> set to "auto", <em>block</em> set to "start",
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -319,12 +319,36 @@ invoke.
 The scroll to text specification proposes an amendment to
 [[html#scroll-to-fragid]]. In summary, if a [=text fragment directive=] is
 present and a match is found in the page, the text fragment takes precedent over
-the element fragment as the indicated part of the document.
+the element fragment as the indicated part of the document. We amend the
+indicated part of the document to possibly include a [[DOM#range|Range]] that is
```
bokand: s/possibly/optionally

index.bs
```diff
@@ -337,9 +361,54 @@ The indicated part of the document</a>.
...
+
+Move the [[cssom-view#scroll-an-element-into-view|scroll an element into
+view]] algorithm's steps 3-14 into a new algorithm <dfn>scroll a DOMRect into
+view</dfn>, with input <a
+href="https://drafts.fxtf.org/geometry-1/#domrect">DOMRect</a> <em>bounding
+border box</em> and
+[[cssom-view#dictdef-scrollintoviewoptions|ScrollIntoViewOptions]] dictionary
+<em>options</em>.
+<div class="note">
+<em>bounding border box</em> is renamed from <em>element bounding border
```
bokand: Lets just call it "bounding box", it would work just as well if you passed in a padding box.

index.bs
```diff
@@ -337,9 +361,54 @@ The indicated part of the document</a>.
...
+### Scroll a DOMRect into view ### {#scroll-rect-into-view}
+<div class="note">
+This section describes a refactoring of the CSSOMVIEW's
+[[cssom-view#scroll-an-element-into-view|scroll an element into view]] algorithm 
+to separate the steps for scrolling a DOMRect into view, so it can be used to
+scroll a Range into view.
+</div>
+
+Move the [[cssom-view#scroll-an-element-into-view|scroll an element into
+view]] algorithm's steps 3-14 into a new algorithm <dfn>scroll a DOMRect into
```
bokand: This means "scroll a DOMRect into view" doesn't have the recursive behavior of scrolling all ancestor scrollers, it probably should? Either you'll need to move that part of "scroll-an-element-into-view" there and add a `startingElement` parameter or you'll need to copy it into "scroll-a-range-into-view"

index.bs

index.bs

index.bs
nickburris: Done, moved the recursive part into scroll a DOMRect into view and added startingElement parameter.

index.bs
```diff
@@ -337,8 +337,8 @@ fragment]] algorithm with the following:
 3. Otherwise:
     3. If <em>range</em> is non-null:
         1. [=scroll a Range into view|Scroll range into view=], with
-            <em>behavior</em> set to "auto", <em>block</em> set to "start", and
-            <em>inline</em> set to "nearest".
+            containingElement <em>target</em>, <em>behavior</em> set to "auto",
+            <em>block</em> set to "start", and <em>inline</em> set to "nearest".
```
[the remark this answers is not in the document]

nickburris:
Thanks!

Issue #73, opened 2019-11-29
Initializing the `TreeWalker`


tilgovi:
From the find a target text algorithm steps (section 2.4.3), step 13 initializes a `TreeWalker` but does not specify what the `root` node or the `whatToFind` filter should be.

tilgovi:
I imagine this would cause an immediate issue with a literal implementation of these steps because the first search would fail the assertion that `walker.currentTreeNode` is a text node.

bzbarsky:
Yeah, none of the treewalker stuff seems to make any sense.  It's created in an invalid state, advanced in ways that make no sense given the state it might be in, etc.  I can't figure out what behavior it's actually trying to aim for....

domenic:
It's also quite strange to use a JavaScript API to walk the tree, instead of walking the underlying spec tree structures directly. What happens if someone does `window.TreeWalker = undefined`, or `TreeWalker.prototype.nextNode = undefined`?

PR #85, opened 2020-02-19
[Spec] Factor out parsing logic

bokand:
This is a small cleanup that makes it a bit easier to follow the logic
for finding a Range from a text directive by splitting out the logic
that parses the string into its individual components.

Also renames `Find a target text` to `Find a matching Range` to
more clearly indicate that the algorithm turns a text directive into
a Range in the Document.

This is a first step in fixing and improving some of the issues in the
tree walking algorithm as part of #73.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -575,14 +575,14 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
   </p>
 </div>
 
-1. Let <var>raw target text</var> be the substring of <em>text directive
+1. Let |raw target text| be the substring of <em>text directive
    input</em> starting at index 5.
    <div class="note">
      This is the remainder of the <em>text directive input</em> following,
      but not including, the "text=" prefix.
    </div>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -549,156 +553,177 @@ scrolled into view.
...
+  <p>
+    Returns null if the input is invalid or fails to parse in any way.
+    Otherwise, returns a <a spec=infra>struct</a> containing four strings:
+    <em>textStart</em>, <em>textEnd</em>, <em>prefix</em>, <em>suffix</em>.
+    Only <em>textStart</em> is required to be non-null as the other parameters
+    are optional and may be omitted in a valid text directive string.
+  </p>
+</div>
+
+1. Let <var>raw target text</var> be the substring of <em>text directive
```
nickburris: nit: for consistency this algorithm should use the `| |` syntax instead of `<var>` (this can just be done in a follow up patch if that's easier)

bokand: Thanks, missed this one. Fixed

PR #86, opened 2020-02-19
Refer to the new HTML user activation model

### LINKED ISSUE #81: Refer to new user activation model
The HTML [user activation model](https://html.spec.whatwg.org/multipage/interaction.html#tracking-user-activation) was [updated](https://github.com/whatwg/html/commit/8f8c1f50158736b3cf16188377a0974a20367c8b) to expand on the concept of "triggered by user activation", which is [used in our draft spec](https://wicg.github.io/ScrollToTextFragment/#allow-text-fragment-directive-flag).

Per the [spec author guidance doc](https://docs.google.com/document/d/14wT89JZ0qeRehXGkcn3_meXxjvlHKgM9d7aJj80kQcQ/edit#heading=h.8c4m9a663yq5), this should be updated to refer to Window's transient activation state.

nickburris:
The HTML user activation model has been updated such that we can more precisely refer to the Window's transient activation state.

Fixes #81.


index.bs
```diff
@@ -421,24 +421,23 @@ asynchronous task to find and set the indicated part of the document.
...
 these steps after step 1:
 
-2. Let <em>is user activated</em> be true if the current navigation was <a
-   href="https://html.spec.whatwg.org/#triggered-by-user-activation"> triggered
-   by user activation</a>
-3. Set <em>document</em>'s <em>allowTextFragmentDirective</em> flag to the
-   result of running [[#should-allow-text-fragment]] with <em>is user
-   activated</em>, <em>incumbentNavigationOrigin</em>, and <em>document</em>.
+2. Let |is user activated| be true if |document|'s associated {{Window}} has
+   <a spec=html>Transient activation</a>.
```
bokand: Hmm, we care that the navigation came from a user gesture - I suspect that's not the same as the window in the navigated-to document having transient activation.

@domenic @mustaqahmed - could you confirm how we should be phrasing this if we want to only allow the API (this happens during document loading) if the navigation comes from a user activation (as opposed to, e.g. script initiated `window.location` changes.)

domenic: Hmm, that's a little tricky as you're persisting state across page loads. This spec does something similar, but doesn't seem to go all the way to what you'd want... https://w3c.github.io/webappsec-fetch-metadata/#request-user-activation-flag

bokand:
Closing this one as I updated this section in #103

PR #90, opened 2020-02-28
[Spec] Re-write text matching steps

### LINKED ISSUE #73: Initializing the `TreeWalker`

bokand:
This commit is a re-write of the text matching portions of the spec.

It avoids using TreeWalker and is more specific and detailed in how the
various algorithms work.

Fixes #73

***

### :boom: Error: write EPROTO 139705468127104:error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version:../deps/openssl/openssl/ssl/s23_clnt.c:772:
 :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 12, 2020, 8:46 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [HTML Diff Service](http://services.w3.org/htmldiff) - The HTML Diff Service is used to create HTML diffs of the spec changes suggested in a pull request.

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2FWICG%2FScrollToTextFragment%2Fpull%2F90%2Fe770c01.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fbokand%2FScrollToTextFragment%2Fpull%2F90.html)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/ScrollToTextFragment%2390.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -171,35 +171,29 @@ create and initialize a Document object</a> steps to parse and remove the
 
 Replace steps 7 and 8 of this algorithm with:
 
-7. Let <em>url</em> be null
-8. If <em>request</em> is non-null, then set <em>document</em>'s
-    [=Document/URL=] to <em>request</em>'s
-    [=request/current URL=].
-9. Otherwise, set <em>url</em> to <em>response</em>'s
-    [=response/URL=].
-10. Let <em>raw fragment</em> be equal to <em>url</em>'s
-    [=url/fragment=].
-11. Let <em>fragment directive position</em> be a
-    [=string/position variable=] that points to the
-    beginning of <em>raw fragment</em>.
-12. While the string starting at position <em>fragment directive position</em>
-    does not start with the [=fragment directive delimiter=] and <em>fragment
-    directive position</em> does not point past the end of <em>raw fragment</em>:
-    1. Advance <em>fragment directive position</em> by 1.
-13. If <em>fragment directive position</em> does not point past the end of
-    <em>raw fragment</em>:
-    1. Let <em>fragment</em> be the substring of <em>raw fragment</em> ending at
-        <em>fragment directive position</em>.
-    2. Advance <em>fragment directive position</em> by the length of [=fragment
-        directive delimiter=].
-    3. Let <em>fragment directive</em> be the substring of <em>raw fragment</em>
-        starting at <em>fragment directive position</em>.
-    4. Set <em>url</em>'s [=url/fragment=] to
-        <em>fragment</em>.
-    5. Set <em>document</em>'s [=fragment directive=] to <em>fragment
-        directive</em>. (Note: this is stored on the document but not
-        web-exposed)
-14. Set <em>document</em>'s [=Document/URL=] to be <em>url</em>.
+7. Let |url| be null
+8. If |request| is non-null, then set |document|'s [=Document/URL=] to
+   |request|'s [=request/current URL=].
+9. Otherwise, set |url| to |response|'s [=response/URL=].
+10. Let |raw fragment| be equal to |url|'s [=url/fragment=].
+11. Let |fragmentDirectivePosition| be an integer initialized to 0.
+12. While the substring of |raw fragment| starting at position
+    |fragmentDirectivePosition| does not begin with the [=fragment directive
+    delimiter=] and
+    |fragmentDirectivePosition| does not point past the end of |raw fragment|:
+    1. Increment |fragmentDirectivePosition| by 1.
+13. If |fragmentDirectivePosition| does not point past the end of |raw
+    fragment|:
+    1. Let |fragment| be the substring of |raw fragment| starting at 0 of count
+       |fragmentDirectivePosition|.
+    1. Advance |fragmentDirectivePosition| by the length of [=fragment
+       directive delimiter=].
+    1. Let |fragment directive| be the substring of |raw fragment| starting at
+       |fragmentDirectivePosition|.
+    1. Set |url|'s [=url/fragment=] to |fragment|.
+    1. Set |document|'s [=fragment directive=] to |fragment directive|. (Note:
+       this is stored on the document but not web-exposed)
+14. Set |document|'s [=Document/URL=] to be |url|.
 
 <div class="note">
   These changes make a URL's fragment end at the [=fragment directive
@@ -226,14 +220,11 @@ run these steps:
   </p>
   <p>
     Returns null if the input is invalid or fails to parse in any way.
-    Otherwise, returns a <a spec=infra>struct</a> containing four strings:
-    |textStart|, |textEnd|, |prefix|, |suffix|.
-    Only |textStart| is required to be non-null as the other parameters
-    are optional and may be omitted in a valid text directive string.
+    Otherwise, returns a [=ParsedTextDirective=].
   </p>
 </div>
 
-1. [=/assert=]: |raw target text| matches the production [=TextDirective=].
+1. [=/Assert=]: |raw target text| matches the production [=TextDirective=].
 1. Let |raw target text| be the substring of |text directive
    input| starting at index 5.
    <div class="note">
@@ -243,25 +234,37 @@ run these steps:
 1. Let |tokens| be a <a for=/>list</a> of strings that is the result of
    <a lt="split on commas">splitting |raw target text| on commas</a>.
 1. If |tokens| has size less than 1 or greater than 4, return null.
-1. If any of |tokens|' items are the empty string, return null.
-1. Let |retVal| be a <a spec=infra>struct</a> consisting of four strings:
-   |textStart|, |textEnd|, |prefix| and |suffix|, each initialized to null.
+1. If any of |tokens|'s items are the empty string, return null.
+1. Let |retVal| be a [=ParsedTextDirective=] with each of its items initialized
+   to null.
 1. Let |potential prefix| be the first item of |tokens|.
 1. If the last character of |potential prefix| is U+002D (-), then:
-    1. Set <em>retVal.prefix</em> to the result of removing the last character from
-       |potential prefix|.
+    1. Set |retVal|'s [=ParsedTextDirective/prefix=] to the result of
+       removing the last character from |potential prefix|.
     1. <a spec=infra for=list>Remove</a> the first item of the list |tokens|.
 1. Let |potential suffix| be the last item of |tokens|, if one exists.
 1. If the first character of |potential suffix| is U+002D (-), then:
-    1. Set <em>retVal.suffix</em> to the result of removing the first character from
-       |potential suffix|.
+    1. Set |retVal|'s [=ParsedTextDirective/suffix=] to the result of
+       removing the first character from |potential suffix|.
     1. <a spec=infra for=list>Remove</a> the last item of the list |tokens|.
 1. If |tokens| has <a spec=infra for=list>size</a> not equal to 1 nor 2 then
    return null.
-1. Let <em>retVal.textStart</em> be the first item of |tokens|.
-1. If |tokens| has <a spec=infra for=list>size</a> 2, then let
-   <em>retVal.textEnd</em> be the last item of |tokens|.
-1. Return |retVal|
+1. Set |retVal|'s [=ParsedTextDirective/textStart=] be the first item of |tokens|.
+1. If |tokens| has <a spec=infra for=list>size</a> 2, then set |retVal|'s
+   [=ParsedTextDirective/textEnd=] be the last item of |tokens|.
+1. Return |retVal|.
+
+A <dfn>ParsedTextDirective</dfn> is a <a spec=infra>struct</a> that consists of
+four strings: <dfn for="ParsedTextDirective">textStart</dfn>,
+<dfn for="ParsedTextDirective">textEnd</dfn>,
+<dfn for="ParsedTextDirective">prefix</dfn>, and
+<dfn for="ParsedTextDirective">suffix</dfn>. [=ParsedTextDirective/textStart=]
+is required to be non-null. The other three items may be set to null,
+indicating they weren't provided. The empty string is not a valid value for any
+of these items.
+
+See [[#syntax]] for the what each of these components means and how they're
+used.
 
 ### Fragment directive grammar ### {#fragment-directive-grammar}
 A <dfn>valid fragment directive</dfn> is a sequence of characters that appears
@@ -428,7 +431,7 @@ to a [=text fragment directive=]-invoking URL, they would be able to determine
 the existence of a text snippet by measuring how long the navigation call takes.
 
 <div class="note">
-  The restrictions in [[#should-allow-text-fragment]] should prevent this
+  The restrictions in [[#restricting-the-text-fragment]] should prevent this
   specific case; in particular, the no-same-document-navigation restriction.
   However, these restrictions are provided as multiple layers of defence.
 </div>
@@ -443,7 +446,10 @@ continue to walk the tree even after a match is found in
 [[#text-directive-to-range]].  Alternatively, it <em>may</em> schedule an
 asynchronous task to find and set the indicated part of the document.
 
-### Should Allow Text Fragment ### {#should-allow-text-fragment}
+### Restricting the Text Fragment ### {#restricting-the-text-fragment}
+
+To determine whether a navigation <dfn>should allow a text fragment</dfn>,
+follow these steps:
 
 <div class="note">
   TODO: This should really only prevent potentially observable side-effects like
@@ -451,23 +457,22 @@ asynchronous task to find and set the indicated part of the document.
   allowed in all cases.
 </div>
 <div class="note">
-  This algorithm has input <em>is user triggered, incumbentNavigationOrigin,
-  document</em> and returns a boolean indicating whether a [=text fragment
+  This algorithm has input |is user triggered|, |incumbentNavigationOrigin|,
+  |document| and returns a boolean indicating whether a [=text fragment
   directive=] should be allowed to invoke on the given Document.
 </div>
 
-1. If <em>is user triggered</em> is false, return false.
-2. If the {{Document}} of the <a spec=HTML>latest entry</a> in
-    <em>document</em>'s [=browsing context=]'s
-    <a spec=HTML>session history</a> is equal to <em>document</em>,
-    return false.
-    <div class="note">
-      i.e. Forbidden on a same-document navigation.
-    </div>
-3. If <em>incumbentNavigationOrigin</em> is equal to the origin of
-   <em>document</em> return true.
-4. If <em>document</em>'s <em>browsingContext</em> is a top level browsing
-   context and its
+1. If |is user triggered| is false, return false.
+1. If the [=document=] of the <a spec=HTML>latest entry</a> in
+   |document|'s [=Document/browsing context=]'s <a spec=HTML>session history</a> is
+   equal to |document|, return false.
+   <div class="note">
+     i.e. Forbidden on a same-document navigation.
+   </div>
+1. If |incumbentNavigationOrigin| is equal to the origin of |document| return
+   true.
+1. If |document|'s [=Document/browsing context=] is a [=top-level browsing
+   context=] and its
    <a href="https://html.spec.whatwg.org/multipage/browsers.html#tlbc-group">group</a>'s
    <a spec=HTML>browsing context set</a> has length 1 return true.
    <div class="note">
@@ -476,9 +481,9 @@ asynchronous task to find and set the indicated part of the document.
      browsing context group to which the navigator does not have script access
      and which may be placed into a separate process.
    </div>
-5. Otherwise, return false.
+1. Otherwise, return false.
 
-### allowTextFragmentDirective flag ### {#allow-text-fragment-directive-flag}
+To set the <dfn>allowTextFragmentDirective</dfn> flag, follow these steps:
 
 <div class="note">
   The algorithm to determine whether or not a text fragment directive should be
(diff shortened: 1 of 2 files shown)


bokand:
@nickburris: It's a big change. I'll see if I can find a spec expert to review later but could you take a first pass?

Put up on https://bokand.github.io/spec.html for ease of review. I've pretty much rewritten all of section "3.5 Navigating to a Text Fragment" so you might just want the published copy here rather than a diff of the markdown.

bokand:
Good idea, added some non-normative notes and an example

bokand:
Thanks for the review. Note: there are still issues even with this update, two that are top of mind for me:

* Using the raw `CharacterData.data` to perform the match (naively) is wrong because this is the raw character data which doesn't necessarily correspond to what's rendered. e.g. whitespace is uncollapsed in this form. (I'll investigate this and fix in a followup)
* The specified word-bounding checks for word boundaries at the time when a string is searched in the concatenated string of all continuous nodes in a block container. This collapses inline nodes. However, Chrome's implementation considers inline nodes to be word bounds. (I think we should change Chrome's implementation to match the spec text).

Will address this and others in follow-ups

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+    “forest ranger”.
+  </div>
+
+  See [[#word-boundaries]] for details and more examples.
 </div>
 
-1. While the input <em>walker.currentNode</em> is not null and
-    <em>walker.currentNode</em> is not a text node:
-    1. Advance the current node by calling {{TreeWalker/nextNode()|walker.nextNode()}}.
+1. [=/assert=]: each item in |nodes| is a {{Text}} node.
```
domenic:
```suggestion
1. [=/Assert=]: each item in |nodes| is a {{Text}} node.
```

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+1. Let |searchBuffer| be the [=string/concatenate|concatenation=] of the
+   [=CharacterData/data=] field of each item in in |nodes|.
```
domenic:
```suggestion
   [=CharacterData/data=] of each item in in |nodes|.
```

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+1. Let |searchStart| be 0
```
domenic:
```suggestion
1. Let |searchStart| be 0.
```

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+1. If the first item in |nodes| is |searchRange|'s [=range/start node=] then
+   set |searchStart| to |searchRange|'s [=range/start offset=].
+1. Let |start|, |end| be [=/boundary point=]s, initially null.
```
domenic:
```suggestion
1. Let |start| and |end| be [=/boundary points=], initially null.
```

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+1. Let |matchIx| be null.
```
domenic: Abbreviating "Index" as "Ix" seems pretty confusing. I'd rather just spell it out.

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+1. While |matchIx| is null
+    1. Let |matchIx| be the first [=string/position variable=] in |searchBuffer|, greater
```
domenic: This doesn't seem like the right use of "position variable". A position variable is a specific "mutable reference" used with "collect a sequence of code points".

I think your "in plain language" version is fine and should replace this whole step.

And, throughout the rest of the specification, "position variable" should be replaced with "integer" or similar.

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+       index=] |endIx| run over |nodes| with |isEnd| true.
+    1. If the substring of |searchBuffer| starting at |matchIx| and of length
+       |queryString|'s [=string/length=] is not [=word bounded=], given the <a
+       spec=html>language</a> from each of |start| and |end|'s [=boundary
+       point/node=]s as the |startLocale| and |endLocale|:
+         1. Let |searchStart| be |matchIx| + 1.
+         1. Set |matchIx| to null.
+1. Let |endInset| be 0.
+1. If the last item in |nodes| is |searchRange|'s [=range/end node=] then set
+   |endInset| to (|searchRange|'s [=range/end node=]'s [=node/length=] -
```
domenic:
```suggestion
   |endInset| to (|searchRange|'s [=range/end node=]'s [=node/length=] &minus;
```

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+   |searchRange|'s [=range/end offset=])
+   <div class="note">
+     |endInset| is the offset from the last position in the last node in the
+     reverse direction. Alternatively, it is the length of the node that's not
+     included in the range.
+   </div>
+1. If |matchIx| + |queryString|'s [=string/length=] is greater than or equal to
+   |searchBuffer|'s length - |endInset| return null.
```
domenic:
```suggestion
   |searchBuffer|'s length &minus; |endInset| return null.
```

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+   <div class="note">
+     If the match runs past the end of the search range, return null.
+   </div>
+1. [=/assert=]: |start| and |end| are non-null, valid [=/boundary point=]s in
```
domenic:
```suggestion
1. [=/Assert=]: |start| and |end| are non-null, valid [=/boundary points=] in
```

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+    than the start of the next node.
+  </p>
 </div>
+
+1. Let |counted| be 0.
+1. For each |curNode| of |nodes|:
+    1. Let |nodeEnd| be |counted| + |curNode|'s [=node/length=].
+    1. If |isEnd| is true, add 1 to |nodeEnd|.
+    1. If |nodeEnd| is greater than |index| then:
+        1. Return the [=/boundary point=] (|curNode|, |index| - |counted|).
```
domenic:
```suggestion
        1. Return the [=/boundary point=] (|curNode|, |index| &minus; |counted|).
```

index.bs
```diff
@@ -791,17 +1036,58 @@ API for word boundary matching.
   alphabet as valid, one-letter words.
```
domenic: The paragraph above this contains "locales without a word-separating character (e.g. space)". This is confusing since it makes it sound like "space" is one of those locales. (Outer space, perhaps? 😉)

index.bs
```diff
@@ -791,17 +1036,58 @@ API for word boundary matching.
...
-        immediately follows |range|, then return |range|.
-2. Return null.
+To determine if a substring of a larger string is <dfn>word bounded</dfn>,
+follow these steps:
+
+<div class="note">
+  This algorithm takes as input a <a spec=infra>string</a> |text|, a
+  [=string/position variable=] |start position| and number |count|; representing
+  a substring in |text|, as well as a |startLocale| and |endLocale| specifying
+  the language of the string at each end of the match.
```
domenic: These inputs should be given in the `<dfn>` "header", not in a non-normative note.

index.bs
```diff
@@ -791,17 +1036,58 @@ API for word boundary matching.
...
+
+  |start position| and |count| are assumed to be valid in that they represent a
+  substring within the bounds of |text|.
+
+  Returns true if the substring is word bounded, false otherwise.
```
domenic: This seems pretty redundant. The name of the algorithm makes this clear.

index.bs
```diff
@@ -791,17 +1036,58 @@ API for word boundary matching.
...
+
+1. Using locale |startLocale|, let |left bound| be the last word boundary in
+   |text| that precedes |start position|th [=code point=] of |text|.
+   <div class="note">
+     A string will always contain at least 2 word boundaries before the first
+     code point and after the last code point of the string.
+   </div>
+1. If the first [=code point=] of |text| following |left bound| is not at
+   position |start position| return false.
+1. Let |end position| be (|start position| + |count| - 1).
```
domenic:
```suggestion
1. Let |end position| be (|start position| + |count| &minus; 1).
```

index.bs
```diff
@@ -498,27 +548,23 @@ the following:
...
-            <em>range</em>; return.
+1. Let |fragment directive string| be the |document|'s [=fragment directive=].
+1. If |document|'s [[#allow-text-fragment-directive-flag]] is true then:
+    1. Let |ranges| be a <a spec=infra>list</a> that is the result of running
+       the [=process a fragment directive=] steps with |fragment directive
+       string| and |document|.
+    1. If |ranges| is non-empty, then:
+        1. Let |range| be the first item of |ranges|.
+           <div class="note">
+             The first {{Range}} in |ranges| is specifically
```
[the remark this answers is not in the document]
domenic: A type and its constructor are equivalent in JavaScript, but you're right they're distinct IDL concepts. Sorry for being confusing. In this case I think I meant the interface.

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
-  </p>
+  This algorithm takes as input four strings: |prefix|, |textStart|, |textEnd|,
+  and |suffix| and a [=Document=] |document| in which to search. It returns a
+  {{Range}} that points to a the first text passage within the document that
+  matches the searched-for text and satisfies the surrounding context. Returns
+  null if no such passage exists.
+
+  See [[#syntax]] for a description of how these parameters are interpreted.
+
+  |textEnd| is optional. If omitted, this is an "exact" search and the returned
```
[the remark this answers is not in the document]
domenic: I haven't looked at the new code yet, but as long as it's a spec algorithm, returning typed structs makes perfect sense. Returning it from a JS method, not so much, but in the spec-layer, it's all good.

I would always stick with `|parsedValues|'s [=ParsedTextDirective/textStart=]`, so that it's closer to the equivalent code.

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+        1. Run the [=find a range from a node list=] steps given |query|,
+           |searchRange|, and |textNodeList|, as input. If the resulting
+           {{Range}} is not null, then return it.
+        1. [=/assert=]: |curNode| [=tree/following|follows=] |searchRange|'s
+           [=range/start node=].
+        1. Set |searchRange|'s [=range/start=] to the [=/boundary point=]
+           (|curNode|, 0).
+1. Return null
+
+A node is <dfn>search invisible</dfn> if it is an {{HTMLElement}} and meets any
```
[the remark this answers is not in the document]
domenic: " just in general that HTMLXxxxElement !== HTMLElement" is what I meant.

In specs "is" generally means "equals". If you want to use "implements", you can use https://heycam.github.io/webidl/#implements (although https://github.com/heycam/webidl/issues/659#issuecomment-467444078 points out that _technically_ you can't use it on Web IDL values... but that's basically just a bug in the definition).

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+of the following conditions:
+1. The [=computed value=] of its 'display' property is 'none'
+1. It has a [=Element/local name=] equal to any of the following: "area",
+   "base", "basefont", "bgsound", "br", "col", "embed", "frame", "hr", "img",
+   "keygen", "link", "menuitem", "meta", "param", "source", "track", "wbr"
+1. Is any of the following types: {{HTMLIFrameElement}}, {{HTMLImageElement}},
+   {{HTMLMeterElement}}, {{HTMLObjectElement}}, {{HTMLProgressElement}},
+   {{HTMLStyleElement}}, {{HTMLScriptElement}}, {{HTMLVideoElement}},
+   {{HTMLAudioElement}}
```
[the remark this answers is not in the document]
domenic: They really should be equivalent. Interfaces are not 1:1 with element names, e.g. both `q` and `blockquote` share `HTMLQuoteElement`, but that doesn't seem to apply to any of the cases being inspected here.

I guess in Chromium it's motivated by code reuse. The spec equivalent of ShouldSerializeEndTag is https://html.spec.whatwg.org/#serializes-as-void. We could use that here, I guess? In which case I am no longer bothered by the mismatch, and you could use interface types here in addition to the https://html.spec.whatwg.org/#serializes-as-void reference. It's just having the two side by side that is weird.

index.bs
```diff
@@ -555,223 +601,422 @@ Define a new algorithm <dfn>scroll a Range into view</dfn>, with input
...
+A node is part of a <dfn>non-searchable subtree</dfn> if it is or has an
+ancestor that is [=search invisible=].
+
+A node is a <dfn>visible text node</dfn> if it is a {{Text}} node, the
+[=computed value=] of its 'visibility' property is 'visible', and it is <a
+spec=html>being rendered</a>.
+
+A node <dfn>has block-level display</dfn> if the [=computed value=] of its
+'display' property is any of 'block', 'table', 'flow-root', 'grid', 'flex',
+'list-item'.
```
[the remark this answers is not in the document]
bokand: From other examples, I've found that e.g. `''display/grid''` should do the trick. They format correctly but they don't link (which is probably ok since the link to disply) except for grid and flex which link to the propdef so I'm guessing there's a bug here somewhere. There's a note in the [bikeshed doc](https://tabatkins.github.io/bikeshed/#:~:text=Note:%20Currently%20the%20CSS%20value%20autolinking%20syntax%20(''foo'')%20does%20not%20work.) that may be relevant.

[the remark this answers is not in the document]
bokand: Actually, never mind, they do work - I think there's some interference with link-defaults I've added in a followup but I'll fix that.

index.bs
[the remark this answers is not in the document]
bokand: sg, I use `|parsedValues|'s [=ParsedTextDirective/textStart=]` in all normative text, there's one or two cases where I'm a bit looser in notes but that should be ok since it's non-normative (and the context is clear)?

index.bs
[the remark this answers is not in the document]
bokand: Ah, got it. Thanks!

index.bs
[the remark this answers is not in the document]
bokand: Done!

bokand:
@domenic: friendly ping to take a look at the changes/replies

(PS, I'm still trying to adjust to the GitHub review system - I'm not sure I'm holding it right - so if there's anything I should/shouldn't be doing for easy reviews please lmk).

domenic:
Thanks for the ping! I'll do my best to get to this tomorrow.

index.bs
```diff
@@ -548,230 +618,420 @@ algorithm with a call to [=scroll a DOMRect into view=]:
...
+       index=] |endIx| run over |nodes| with |isEnd| true.
+    1. If the substring of |searchBuffer| starting at |matchIndex| and of length
+       |queryString|'s [=string/length=] is not [=word bounded=], given the <a
+       spec=html>language</a> from each of |start| and |end|'s [=boundary
+       point/nodes=] as the |startLocale| and |endLocale|:
+         1. Let |searchStart| be |matchIndex| + 1.
+         1. Set |matchIndex| to null.
+1. Let |endInset| be 0.
+1. If the last item in |nodes| is |searchRange|'s [=range/end node=] then set
+   |endInset| to (|searchRange|'s [=range/end node=]'s [=node/length=] &minus
```
domenic:
```suggestion
   |endInset| to (|searchRange|'s [=range/end node=]'s [=node/length=] &minus;
```

domenic: This problem occurs throughout the document.

index.bs
```diff
@@ -784,24 +1044,68 @@ API for word boundary matching.
...
-        |text| after |range|.
-    5. If |left bound| immediately precedes |range| and |right bound|
-        immediately follows |range|, then return |range|.
-2. Return null.
+To determine if a substring of a larger string is <dfn>word bounded</dfn>,
+given a <a spec=infra>string</a> |text|, an integer |startPosition|, number
+|count|, and locales |startLocale| and |endLocale|, follow these steps:
+
+|startLocale| and |endLocale| must be a valid BCP 47[<a
+href="https://tools.ietf.org/html/bcp47">BCP47</a>] language tag, or the empty
```
domenic: It'd be better to do this as `[[BCP47]]`. Then it will end up in your bibliography. (In general, raw `<a>` elements are somewhat of a smell in Bikeshed documents.)

index.bs
```diff
@@ -213,6 +212,65 @@ the fragment is the string "test" and the [=fragment directive=] is the string
...
+1. If |tokens| has size less than 1 or greater than 4, return null.
+1. If any of |tokens|'s items are the empty string, return null.
+1. Let |retVal| be a [=ParsedTextDirective=] with each of its items initialized
+   to null.
+1. Let |potential prefix| be the first item of |tokens|.
+1. If the last character of |potential prefix| is U+002D (-), then:
+    1. Set |retVal|'s [=ParsedTextDirective/prefix=] to the result of
+       removing the last character from |potential prefix|.
+    1. <a spec=infra for=list>Remove</a> the first item of the list |tokens|.
+1. Let |potential suffix| be the last item of |tokens|, if one exists.
```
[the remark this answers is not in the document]
bokand: Done (`potential prefix` cannot be null though since we check above that `tokens` has at least 1 item. `potential suffix` has the issue because we might have already removed the only item in the prefix case)

index.bs
```diff
@@ -213,6 +212,65 @@ the fragment is the string "test" and the [=fragment directive=] is the string
...
+       removing the first character from |potential suffix|.
+    1. <a spec=infra for=list>Remove</a> the last item of the list |tokens|.
+1. If |tokens| has <a spec=infra for=list>size</a> not equal to 1 nor 2 then
+   return null.
+1. Set |retVal|'s [=ParsedTextDirective/textStart=] be the first item of |tokens|.
+1. If |tokens| has <a spec=infra for=list>size</a> 2, then set |retVal|'s
+   [=ParsedTextDirective/textEnd=] be the last item of |tokens|.
+1. Return |retVal|.
+
+A <dfn>ParsedTextDirective</dfn> is a <a spec=infra>struct</a> that consists of
```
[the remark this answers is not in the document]
bokand: Thanks!

index.bs
```diff
@@ -390,34 +448,36 @@ has been successfully found</em>.
...
 continue to walk the tree even after a match is found in
-[[#find-a-matching-range]].  Alternatively, it <em>may</em> schedule an
+[[#text-directive-to-range]].  Alternatively, it <em>may</em> schedule an
 asynchronous task to find and set the indicated part of the document.
 
-### Should Allow Text Fragment ### {#should-allow-text-fragment}
+### Restricting the Text Fragment ### {#restricting-the-text-fragment}
+
+To determine whether a navigation <dfn>should allow a text fragment</dfn>,
+follow these steps:
```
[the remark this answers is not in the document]
bokand: Done

index.bs
```diff
@@ -498,27 +558,37 @@ the following:
...
+        1. Let |range| be the first item of |ranges|.
+           <div class="note">
+             The first [=range=] in |ranges| is specifically
+             scrolled into view. This [=range=], along with the
+             remaining |ranges| should be visually indicated in a way that
+             is not revealed to script, which is left as UA-defined behavior.
+           </div>
+        1. Let |node| be the [=first common ancestor=] of |range|'s
+            [=range/start node=] and [=range/start node=].
+        1. While |node| is not an [=Element=], set |node| to |node|'s
```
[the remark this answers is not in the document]
bokand: Done

index.bs
```diff
@@ -548,230 +618,420 @@ algorithm with a call to [=scroll a DOMRect into view=]:
...
+  directive=] steps are the high level API provided by this section. These
+  return a <a spec=infra>list</a> of [=ranges=] that were matched by the
+  individual directive matching steps, in the order the directives were
+  specified in the fragment directive string.
+
+  If a directive was not matched, it does not add an item to the returned
+  list.
+</div>
+
+To <dfn>process a fragment directive</dfn>, run these steps:
```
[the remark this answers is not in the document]
bokand: Done

index.bs
```diff
@@ -548,230 +618,420 @@ algorithm with a call to [=scroll a DOMRect into view=]:
...
-1. If |end position| is null, then return null.
-1. Advance |end position| by the length of |textEnd|.
-1. Let |match| be a {{Range}} in <em>walker.currentNode</em> with start
-   position |potential start position| and length equal to |end position| - |start
-   position|.
-1. Return |match|.
-
-### Find an exact match with context ### {#find-match-with-context}
+1. Let |searchRange| be a [=range=] with [=range/start=] (|document|, 0) and
+   [=range/end=] (|document|, |document|'s [=node/length=])
```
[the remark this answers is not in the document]
bokand: Oops, it's case-sensitive. Changed to [Node/length](https://dom.spec.whatwg.org/#concept-node-length) here and other instances.

index.bs
```diff
@@ -548,230 +618,420 @@ algorithm with a call to [=scroll a DOMRect into view=]:
...
+             non-whitespace position is at the end of the document.
+           </div>
+        1. [=/Assert=]: |searchRange|'s [=range/start node=] is a {{Text}} node.
+           <div class="note">
+             |searchRange|'s [=range/start=] now points to the next
+             non-whitespace text data following a matched prefix.
+           </div>
+        1. Set |potentialMatch| to the result of running the [=find a string in
+           range=] steps given |parsedValues|'s
+           [=ParsedTextFragment/textStart=] and [=ParsedTextFragment/textEnd=],
```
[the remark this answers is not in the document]
bokand: D'oh! Should be ParsedTextDirective. Fixed.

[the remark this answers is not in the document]
bokand: (I have a followup CL that fixes the formatting errors so I can put this through bikeshed with warnings/errors. This is fixed there but I didn't backport it here until now).

index.bs
bokand: Fixed all instances.

index.bs
bokand: Nice, I tried a few things but didn't find one so fellback to the raw link. Is there a listing somewhere of all references? (I vaguely remember some kind of directory for specs but it was a very long time ago).

domenic: https://www.specref.org/ :)

index.bs
```diff
@@ -548,230 +618,420 @@ algorithm with a call to [=scroll a DOMRect into view=]:
...
+           and |searchRange|.
+           <div class="note">
+             TODO: [=find a string in range=] doesn't take two strings!
```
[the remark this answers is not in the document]
bokand: Two strings and a range :)

Latest commit fixes this (and found another bug where we advance the searchRange too far when the prefix doesn't match: `data:text/html,a a a b b b#:~:text=a a-,b`)

index.bs
bokand: That's the one, thanks!

bokand:
Latest commit has all the latest feedback incorporated.

bokand:
Thanks for the review and informative tips!

PR #95, opened 2020-03-12
[Spec] Explicitly mention 'focus' based attack

bokand:
Another attack vector that's been brought to our attention is using the
fact that fragments (text or otherwise) cause focus to be applied. An
attacker can listen for notifications about losing focus in its own
document and infer that a fragment was successfully matched in a
cross-origin victim.

Our existing security measures mitigate this already, but this commit
adds some text to the non-normative motivation section to make readers
explicitly aware of this technique.

***

### :boom: Error: write EPROTO 139705468127104:error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version:../deps/openssl/openssl/ssl/s23_clnt.c:772:
 :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 12, 2020, 9:25 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [HTML Diff Service](http://services.w3.org/htmldiff) - The HTML Diff Service is used to create HTML diffs of the spec changes suggested in a pull request.

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2FWICG%2FScrollToTextFragment%2Fpull%2F95%2Fbb94206.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fbokand%2FScrollToTextFragment%2Fpull%2F95.html)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/ScrollToTextFragment%2395.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -440,6 +440,14 @@ detectable and distinguished from natural user scrolls.
 </div>
 
 <div class="example">
+  A malicious page embeds a cross-origin victim in an iframe. The victim page
+  contains information sensitive to the user. The malicious page navigates the
+  victim to a text fragment. Since a successful fragment match will cause
+  focus, the malicious page can determine if the text appears in the victim by
+  listening for a blur event in its own document.
+</div>
+
+<div class="example">
   An attacker sends a link to a victim, sending them to a page that displays
   a private token. The attacker asks the victim to read back the token. Using
   a text fragment, the attacker gets the page to load for the victim such that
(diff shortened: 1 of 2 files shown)


bokand:
PR Preview still seems to bug out, but if you expand the error there's a "Related link" that shows the HTML diff.

This PR is simple enough that you don't need it but useful to know for future reviews.

PR #102, opened 2020-05-21
Modify spec to allow text fragment when initiated from browser UI

### LINKED ISSUE #101: [Spec] Needs to consider the case where incumbentNavigationOrigin is null
The [Restricting the Text Fragment](https://wicg.github.io/ScrollToTextFragment/#restricting-the-text-fragment) section checks the incumbentNavigationOrigin to determine if the navigation is cross-origin and should be blocked.

However, if the navigation comes from browser UI (e.g. typing the URL into the address bar, bookmark) we will not have an incumbentNavigationOrigin.

This case should be allowed though so it should be explicitly specified.

bokand:
Fixes #101

***
<a href="https://pr-preview.s3.amazonaws.com/bokand/ScrollToTextFragment/pull/102.html" title="Last updated on Jun 29, 2020, 7:58 PM UTC (7f1e8a2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/scroll-to-text-fragment/102/c09a231...bokand:7f1e8a2.html" title="Last updated on Jun 29, 2020, 7:58 PM UTC (7f1e8a2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -533,9 +533,28 @@ given as input a boolean |is user triggered|, an [=origin=]
 
 1. If |incumbentNavigationOrigin| is null, return true.
     <div class="note">
-      If a navigation originates from browser UI, it's always ok to allow it
-      since it'll be user triggered and the page/script isn't providing the
-      text snippet.
+      <p>
+        If a navigation originates from browser UI, it's always ok to allow it
+        since it'll be user triggered and the page/script isn't providing the
+        text snippet.
+      </p>
+      <p>
+        Note: Depending on the UA, there may be cases where the
+        |incumbentNavigationOrigin| is null but it's not clear that the
+        navigation should be considered as initiated from browser UI. E.g. an
+        "open in new window" context menu item when right clicking on a link.
+        The intent in this item is to distinguish cases where the app/page is
+        able to set the URL from those that are fully under the user's control.
+        In the former we want to prevent activation of the text fragment unless
+        the destination is loaded in a separate browsing context group (so that
+        the source cannot both control teh text snippet and observe
+        side-effects in the navigation).
+      </p>
+      <p>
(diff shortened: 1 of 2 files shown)


bokand:
Hey @arturjanc! Could you just gut-check that this is ok? It comes from [feedback](https://github.com/mozilla/standards-positions/issues/194#:~:text=representing%20a%20navigation%20coming%20from%20the%20browser%20UI) from Mozilla's standards-position.

I think that if the navigation is coming from browser UI then even if an attacker could listen to changes on the page it's ok to scroll/activate because they can't control the text snippet used. Is this right?

arturjanc:
This seems reasonable to me overall, but in the [Fetch Metadata](https://w3c.github.io/webappsec-fetch-metadata/#directly-user-initiated) spec we ran into a lot of nuance when trying to decide which navigations are coming from the browser UI (because we needed to set `Sec-Fetch-Site: none` for browser-controlled navigations, as opposed to `Sec-Fetch-Site: cross-origin` for web-controlled navigations).

The difficulty is that there are a lot of types of navigations in the grey area, e.g. the _Open in new tab_ option in the context menu or a Ctrl+click on a link. In this case the website is under the control of the navigation URL, but the interaction isn't spoofable in the same way as a regular link click or interactionless `window.location` assignment is.

So my guess is that you may at some point need to define this more precisely, and it could make sense to align this with the logic in Fetch Metadata.

bokand:
Thanks Artur, the `Sec-Fetch-Site` header does appear to have a very similar intent. I've added a note about this. (for practical purposes I don't think it matters quite as much here since we're determining whether to check for an isolated browsing context group but the grey areas like "open in new tab/window" will force that to happen anyway.

PR #133, opened 2020-08-12
Explainer and spec enabling client-redirects

### LINKED ISSUE #121: Client-side redirects (e.g. Twitter t.co) loses user gesture, breaks text fragment
I'm not sure of the exact mechanism, but tweeting or sharing a scroll to text fragment link via Twitter or Facebook does not work. As far as I can see the link expands correctly, but chrome does not scroll or highlight. 
eg https://twitter.com/kevinmarks/status/1273197878354247681 contains http://microformats.org/wiki/process#:~:text=property%20name but that doesn't work when clicked, even though https://t.co/SZ7LTv4lUY?amp=1 resolves to it.
Is the 'hide things from the client page' paranoia in section 3.4 causing this?

bokand:
This adds spec and explainer text for how we solve the client-redirect issue.

Fixes #121

***
<a href="https://pr-preview.s3.amazonaws.com/bokand/ScrollToTextFragment/pull/133.html" title="Last updated on Aug 28, 2020, 6:30 PM UTC (27992ad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/scroll-to-text-fragment/133/c897baa...bokand:27992ad.html" title="Last updated on Aug 28, 2020, 6:30 PM UTC (27992ad)">Diff</a>


bokand:
@mikewest PTAL

bokand:
@mikewest mind giving this a quick look over as well?

bokand:
(particularly the spec text in index.bs)

index.bs
```diff
@@ -544,16 +598,16 @@ given as input a boolean |is user triggered|, an [=origin=]
...
         the destination is loaded in a separate browsing context group (so that
-        the source cannot both control teh text snippet and observe
+        the source cannot both control the text snippet and observe
         side-effects in the navigation).
       </p>
       <p>
         TODO: This seems to be very similar to <a
         href="https://w3c.github.io/webappsec-fetch-metadata/#directly-user-initiated">sec-fetch-site</a>
-        so we may wish to integrate this with how that's specified.
+        so should probably integrate with how that's specified once it's ready.
```
mikewest, line 607: We're shipping `Sec-Fetch-Site`, Firefox has implemented it, and I hope it's reasonably specified. What more would you like to see?

index.bs
```diff
@@ -590,18 +644,30 @@ these steps after step 1:
```
mikewest, line 644: You probably should integrate these steps into the "[create and initialize a `Document` object](https://html.spec.whatwg.org/multipage/browsing-the-web.html#initialise-the-document-object)" algorithm, rather than the wrapper.

index.bs
```diff
@@ -590,18 +644,30 @@ these steps after step 1:
...
       <a href="https://w3c.github.io/webappsec-fetch-metadata/#request-user-activation-flag">Sec-Fetch-User</a>.
-      Perhaps we could generalize that for use here and elsewhere?
     </div>
 
-3. Set |document|'s [=allowTextFragmentDirective=] flag to the result of
-    running [=should allow a text fragment=] with |is user activated|,
-    |incumbentNavigationOrigin|, and the document.
+3. If either of |is user activated| or the [=request/textFragmentToken=] flag of
+    |navigationParam|'s |request| object is true, set the |document|'s
+    [=document/textFragmentToken=] flag to true. Otherwise, set it to false. Set the
```
mikewest, line 657: It's probably reasonable to set this flag only for top-level documents? It's not clear to me whether there's a race condition that could allow spawning multiple iframes which would copy the boolean into each, thus allowing several of those frames to cause a text-fragment navigation using their own copy of the flag.

mikewest: (The last sentence aims to address this, I know, but it relies on some sort of pass-by-reference usage of `|navigationParams|`, which is somewhat hard to specify in prose.)

mikewest (commented):
Looks pretty reasonable, thank you.

index.bs
bokand: Thanks, done.

index.bs
bokand: Done and removed the write to `navigationParams`.

Question though: is there a way to reuse the request by spawning an iframe? I'm not quite sure I understand the scenario you described above.

index.bs
```diff
@@ -518,90 +530,181 @@ to find and set the indicated part of the document.
...
+    1. If the [=document=] of the <a spec=HTML>latest entry</a> in
+        |document|'s [=Document/browsing context=]'s <a spec=HTML>session history</a> is
+        equal to |document|, set [=document/allowTextFragmentDirective=] to false
+        and abort these sub-steps.
+        <div class="note">
+          i.e. Forbidden on a same-document navigation.
+        </div>
+    1. If the |navigationParam|'s
+        <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params-request">request</a>
+        has a <a href="https://w3c.github.io/webappsec-fetch-metadata/#http-headerdef-sec-fetch-site">set-fetch-site</a>
```
bokand, line 685: I couldn't get autolinking to the fetch-metadata spec work out using [=sec-fetch-site=], off-hand, any idea what I'm doing wrong?

bokand:
Ping @mikewest, mind taking a second pass?

index.bs
mikewest: It's not a `dfn`, but an `http-header` type, which doesn't (I think?) have a shorthand. You'll need something like ```<a type="http-header">`Sec-Fetch-Site`</a>``` to make this work.

mikewest (approved):
LGTM. Thanks for taking another pass.

index.bs
bokand: Hmm, still having trouble but don't want to get caught up on it - I'll fiddle with it later.

Issue #135, opened 2020-08-19
TypeScript implementation & polyfill


Treora:
Hi all, sorry to open an issue for what is rather an introduction&announcement (but where else to share this?). I have been following this proposal as I have long been wanting the ability to link to arbitrary selections in web pages. A few years ago I made a very similar attempt called [quoteurl][] (my proposed syntax was `#(prefix)"startText...endText"(suffix)`), and later made a browser extension ([Precise links][]) that implements the suggestion from the Web Annotation [selectors&states note][] (i.e. the syntax `#selector(type=TextQuoteSelector,exact=blabla…)`).

Now that this proposal has significantly more traction than any before it, as well as a much further worked-out specification, I figured it may be worth jumping on this bandwagon, and try to help make this feature part of the web at large. To better understand the details of the proposal and enable people to play with it in practice, I implemented most of the current draft spec, line by line, in TypeScript:

<https://code.treora.com/gerben/text-fragments-ts>

The implementation is found in [src/index.ts][]. The repo includes a polyfill ([src/polyfill.ts][]) and a small demo, which you can also play with [here][].

For your convenience I also mirror the repo on GitHub, so you can create issues and pull requests [there][] (because distributed version control unfortunately still lacks distributed collaboration tools…).

While coding, I encountered small unclarities and mistakes in the spec, which are noted in the source code in comments marked with with `XXX`. I will open issues and pull requests for some of those, as well as for a few questions about the functionality.

[quoteurl]: https://github.com/Treora/quoteurl
[selectors&states note]: https://www.w3.org/TR/2017/NOTE-selectors-states-20170223/#frags
[Precise links]: https://addons.mozilla.org/en-US/firefox/addon/precise-links/
[src/index.ts]: https://code.treora.com/gerben/text-fragments-ts/src/branch/develop/src/index.ts
[src/polyfill.ts]: https://code.treora.com/gerben/text-fragments-ts/src/branch/develop/src/polyfill.ts
[demo of quoteurl]: https://rawgit.com/Treora/quoteurl/master/example.html#%22Integer%20nisl%20elit%22
[here]: https://temp.treora.com/text-fragments-ts/demo.html
[there]: https://github.com/Treora/text-fragments-ts

Treora:
[...]

PR #136, opened 2020-08-19
Some suggestions and corrections.

Treora:
Following up on #135, some simple suggestions to the spec, each corresponding to an ‘`XXX`’ comment in my implementation’s [index.ts](https://code.treora.com/gerben/text-fragments-ts/src/branch/develop/src/index.ts) (but not *every* of those comments is resolved here).

***
<a href="https://pr-preview.s3.amazonaws.com/Treora/scroll-to-text-fragment/pull/136.html" title="Last updated on Aug 28, 2020, 10:16 PM UTC (9f0b391)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/scroll-to-text-fragment/136/74ba764...Treora:9f0b391.html" title="Last updated on Aug 28, 2020, 10:16 PM UTC (9f0b391)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -127,9 +127,9 @@ exact text "an example text fragment" is the target text.
 
 If the textEnd parameter is also specified, then the text directive refers to a
 range of text in the page. The target text range is the text range starting at
-the first instance of startText, until the first instance of endText that
-appears after startText. This is equivalent to specifying the entire text range
-in the startText parameter, but allows the URL to avoid being bloated with a
+the first instance of textStart, until the first instance of textEnd that
+appears after textStart. This is equivalent to specifying the entire text range
+in the textStart parameter, but allows the URL to avoid being bloated with a
 long text directive.
 
 <div class="example">
@@ -217,24 +217,25 @@ Replace steps 7 and 8 of this algorithm with:
     |request|'s [=request/current URL=].
 9. Otherwise, set |url| to <var ignore=''>response</var>'s [=response/URL=].
 10. Let |raw fragment| be equal to |url|'s [=url/fragment=].
-11. Let |fragmentDirectivePosition| be an integer initialized to 0.
-12. While the substring of |raw fragment| starting at position
-    |fragmentDirectivePosition| does not begin with the [=fragment directive
-    delimiter=] and
-    |fragmentDirectivePosition| does not point past the end of |raw fragment|:
-    1. Increment |fragmentDirectivePosition| by 1.
-13. If |fragmentDirectivePosition| does not point past the end of |raw
-    fragment|:
-    1. Let |fragment| be the substring of |raw fragment| starting at 0 of count
-        |fragmentDirectivePosition|.
-    1. Advance |fragmentDirectivePosition| by the length of [=fragment
-        directive delimiter=].
-    1. Let |fragment directive| be the substring of |raw fragment| starting at
-        |fragmentDirectivePosition|.
-    1. Set |url|'s [=url/fragment=] to |fragment|.
-    1. Set |document|'s [=fragment directive=] to |fragment directive|. (Note:
-        this is stored on the document but not web-exposed)
-14. Set |document|'s [=Document/URL=] to be |url|.
+11. If |raw fragment| is non-null:
+    1. Let |fragmentDirectivePosition| be an integer initialized to 0.
+    2. While the substring of |raw fragment| starting at position
+        |fragmentDirectivePosition| does not begin with the [=fragment directive
+        delimiter=] and |fragmentDirectivePosition| does not point past the end
+        of |raw fragment|:
+        1. Increment |fragmentDirectivePosition| by 1.
+    3. If |fragmentDirectivePosition| does not point past the end of |raw
+        fragment|:
+        1. Let |fragment| be the substring of |raw fragment| starting at 0 of
+            count |fragmentDirectivePosition|.
+        1. Advance |fragmentDirectivePosition| by the length of [=fragment
+            directive delimiter=].
+        1. Let |fragment directive| be the substring of |raw fragment| starting
+            at |fragmentDirectivePosition|.
+        1. Set |url|'s [=url/fragment=] to |fragment|.
+        1. Set |document|'s [=fragment directive=] to |fragment directive|.
+            (Note: this is stored on the document but not web-exposed)
+12. Set |document|'s [=Document/URL=] to be |url|.
 
 <div class="note">
   These changes make a URL's fragment end at the [=fragment directive
@@ -248,8 +249,8 @@ the fragment is the string "test" and the [=fragment directive=] is the string
 "text=foo".
 </div>
 
-To <dfn>parse a text directive</dfn>, on a <a spec="infra">string</a> |textDirectiveString|,
-run these steps:
+To <dfn>parse a text directive</dfn>, on a <a spec="infra">string</a> |text
+directive input|, run these steps:
 
 <div class="note">
   <p>
@@ -265,7 +266,7 @@ run these steps:
   </p>
 </div>
 
-1. [=/Assert=]: |textDirectiveString| matches the production [=TextDirective=].
+1. [=/Assert=]: |text directive input| matches the production [=TextDirective=].
 1. Let |textDirectiveString| be the substring of |text directive
     input| starting at index 5.
     <div class="note">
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -268,21 +269,23 @@ run these steps:
...
     1. <a spec=infra for=list>Remove</a> the last item of the list |tokens|.
 1. If |tokens| has <a spec=infra for=list>size</a> not equal to 1 nor 2 then
     return null.
-1. Set |retVal|'s [=ParsedTextDirective/textStart=] be the first item of |tokens|.
+1. Set |retVal|'s [=ParsedTextDirective/textStart=] be the percent-decoding of
+    the first item of |tokens|.
 1. If |tokens| has <a spec=infra for=list>size</a> 2, then set |retVal|'s
-    [=ParsedTextDirective/textEnd=] be the last item of |tokens|.
+    [=ParsedTextDirective/textEnd=] be the percent-decoding of the last item of
+    |tokens|.
```
Treora, line 288: I suppose these mentions of “percent-decoding” should be linked to <https://url.spec.whatwg.org/#string-percent-decode>, but I’m not sure how as I am new to bikeshed (i.e. the tool, I think I’ve done plenty bikeshedding in my life :)).

index.bs
```diff
@@ -1043,28 +1037,29 @@ meets any of the following conditions:
...
-A node is part of a <dfn>non-searchable subtree</dfn> if it is or has an
-ancestor that is [=search invisible=].
+A node is part of a <dfn>non-searchable subtree</dfn> if it is or has a
+[=shadow-including ancestor=] that is [=search invisible=].
 
 A node is a <dfn>visible text node</dfn> if it is a {{Text}} node, the
-[=computed value=] of its 'visibility' property is ''visibility/visible'', and
-it is <a spec=html>being rendered</a>.
+[=computed value=] of its [=parent element=]'s 'visibility' property is
+''visibility/visible'', and it is <a spec=html>being rendered</a>.
```
Treora, line 1045: Not sure about this one; but I thought the visibility property would not apply to non-element nodes. Is checking its parent element then the solution?

bokand:
Thanks for this! I just got back from vacation and am still catching up but will take a look in the next day or two.

index.bs
```diff
@@ -1004,37 +999,36 @@ in a given [=range=] |range|, run these steps:
         1. [=iteration/Continue=].
     1. If |curNode| is not a [=visible text node=]:
         1. Set |searchRange|'s [=range/start node=] to the next node, in
-            [=shadow-including tree order=].
+            [=shadow-including tree order=], that is not a [=dom/doctype=], and
```
bokand, line 1002: `[=dom/doctype=]` fails to autolink, use `[=doctype=]`.

Would it be better to instead add doctype nodes to the set of [search invisible](https://wicg.github.io/scroll-to-text-fragment/#search-invisible) nodes so that it's skipped in the step above?

index.bs
```diff
@@ -1004,37 +999,36 @@ in a given [=range=] |range|, run these steps:
...
-            [=range=] is not null, then return it.
-        1. [=/Assert=]: |curNode| [=tree/following|follows=] |searchRange|'s
-            [=range/start node=].
-        1. Set |searchRange|'s [=range/start=] to the [=/boundary point=]
-            (|curNode|, 0).
+    1. Let |blockAncestor| be the [=nearest block ancestor=] of |curNode|.
+    1. Let |textNodeList| be a <a spec=infra>list</a> of {{Text}} nodes,
+        initially empty.
+    1. While |curNode| is a [=shadow-including descendant=] of |blockAncestor|
+        and position of the [=/boundary point=] (|curNode|, 0) is not [=boundary
```
bokand, line 1009: > ...|blockAncestor| and position of the...

Should be ...|blockAncestor| and _the_ position of the...

index.bs
```diff
@@ -1004,37 +999,36 @@ in a given [=range=] |range|, run these steps:
...
+    1. If |curNode| is null, then [=iteration/break=].
+    1. [=/Assert=]: |curNode| [=tree/following|follows=] |searchRange|'s
+        [=range/start node=].
+    1. Set |searchRange|'s [=range/start=] to the [=/boundary point=] (|curNode|,
+        0).
 1. Return null.
 
-A node is <dfn>search invisible</dfn> if it is in the [=HTML namespace=] and
-meets any of the following conditions:
+A node is <dfn>search invisible</dfn> if it is an [=element=] is in the [=HTML
```
bokand, line 1030: > if it is an [=element=] is in the [=HTML namespace=]

Should be

> if it is an [=element=] in the [=HTML namespace=]

Does this need to be limited to elements though? My suggestion above about putting doctypes in here would require this to be nodes.

index.bs
bokand: Yes, I believe you are right. AFAICT CSS styles apply only to elements.

index.bs
```diff
@@ -1043,28 +1037,29 @@ meets any of the following conditions:
...
+    1. If |curNode| is not a {{Text}} node and it [=has block-level display=] then
+        return |curNode|.
+    1. Otherwise, set |curNode| to |curNode|'s [=tree/parent=].
 1. Return |node|'s [=Node/node document=]'s [=document element=].
 
-To <dfn>find a range from a node list</dfn> given a search string
-|queryString|, a [=range=] |searchRange|, and a [=/list=] of [=nodes=] |nodes|,
-follow the steps
+To <dfn>find a range from a node list</dfn> given a search string |queryString|,
+a [=range=] |searchRange|, [=/list=] of {{Text}} nodes |nodes|, follow the
```
bokand, line 1061: > |searchRange|, [=/list=] of [=nodes=] |nodes|,

nit: Keep the "and a" before list of nodes:

> |searchRange|, and a [=/list=] of [=nodes=] |nodes|,

index.bs
```diff
@@ -1119,7 +1114,7 @@ follow the steps
       reverse direction. Alternatively, it is the length of the node that's not
       included in the range.
     </div>
-1. If |matchIndex| + |queryString|'s [=string/length=] is greater than or equal to
```
bokand, line 1122: Good catch!

index.bs
```diff
@@ -629,27 +630,29 @@ may be scrolled into view instead of the containing element.
...
-          <div class="note">
-              This otherwise case is the same as the current step 3.3.
-          </div>
+
+3. <a href="https://w3c.github.io/webappsec-permissions-policy/document-policy.html#algo-get-policy-value">Get
+    the policy value</a> for `force-load-at-top` in the
+    [=Document=]. If the result is true, abort these steps.
+4. If <em>range</em> is non-null:
+    1. If the UA supports scrolling of text fragments on navigation, invoke
+        [=scroll a Range into view|Scroll range into view=], with range
```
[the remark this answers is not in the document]
Treora: In `with range <em>range</em>`, the first “range” refers to the parameter name of the algorithm, the second refers to the local variable, so neither specifies the type. I’d think the type of parameters should only be mentioned at the algorithm definition, not during its invocation (the type of the local variable *range* is already stated before). So `[=range=]` would be awkward here.

However, the parameter names could be left out completely, relying on the parameter order: `invoke … with <em>range</em>, <em>target</em>, …`. However I am not sure how to deal with the options dictionary that this algorithm expects; it currently looks like the dictionary is made ‘transparent’ (I mean that the options inside the dictionary are treated as if they were parameters to the algorithm itself). I don’t know what the usual practice is for passing a dictionary.

index.bs
```diff
@@ -574,8 +577,6 @@ given as input a boolean |is user triggered|, an [=origin=]
     </div>
 1. Otherwise, return false.
 
-To set the <dfn>allowTextFragmentDirective</dfn> flag, follow these steps:
```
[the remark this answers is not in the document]
bokand: Just merged #133 so I think you can just rebase over that.

index.bs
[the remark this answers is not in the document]
bokand: I'm somewhat confused since the [scroll an element into view](https://drafts.csswg.org/cssom-view-1/#scroll-an-element-into-view) algorithm doesn't take an the dictionary so this doesn't really make sense anyway. It appears that part of the CSSOM-View spec was recently modified so perhaps that changed? It takes the individual pieces of the dictionary.

Would you mind updating it so that `scroll a DOMRect into view` and `scroll a range into view` both take individual _behavior_, _block_, _inline_ params?

index.bs
bokand: Yep, you're right.

index.bs
bokand: sounds good thanks.

index.bs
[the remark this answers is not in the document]
Treora: >  It appears that part of the CSSOM-View spec was recently modified so perhaps that changed? It now takes the individual pieces of the dictionary.

Three years ago, [apparently](https://github.com/w3c/csswg-drafts/commit/232957f084dd40e63df136fe1c7f5b42ac642002#diff-3c17f0e43c20f8ecf89419d49e7ef5e0L1302).

Might this actually be an argument to keep parameter names in place, as it helps detect spec mismatches? (shouldn’t references to draft or ‘living’ specs perhaps mention a version/date?)

I started updating their signatures, but noticed more mismatches or inaccuracies. E.g. `startingElement` should probably just be `element`; also the steps that are to be moved use the “element width”, which I suppose should become “bounding box’s width”. Might it be worthwhile to just include the steps into this spec, while rewriting them accordingly? (also renaming “element edge A” with “bounding box edge A” or “domrect edge A”, etc.) Or perhaps even to suggest the cssom-view to refactor the algorithm?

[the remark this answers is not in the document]
bokand: Re: keeping parameter names, I don't feel terribly strongly so I'll leave it up to you.

The idea here is that (once we can get another vendor to agree) we can move all these monkeypatches into the appropriate real specs so it shouldn't be too long lived. I guess my preference would be to not copy text into here and keep it just a diff; copying it here wouldn't solve the underlying issue of the source spec changing and would make it harder to tell what the desired changes are when we do go to merge.

This piece in particular seems fairly independent of any actual fragment related changes so might make sense to open a PR against CSSOM-View. I may do that if I can find some time next week (or if you have the cycles, feel free as well).

Issue #137, opened 2020-08-20
Why only match at word boundaries?


Treora:
The current draft specification only allows text matches to occur at word boundaries (see [section §3.5.3](https://wicg.github.io/scroll-to-text-fragment/#word-boundaries)). As clarified in [example 12](https://wicg.github.io/scroll-to-text-fragment/#example-1b2a6f00):

>  The substring "mountain range" is word bounded within the string "An impressive mountain range" but not within "An impressive mountain ranger".

As for the reason, the only hint in the spec is at the top of the section:

> Limiting matching to word boundaries is one of the mitigations to limit cross-origin information leakage.

Given the various other mitigations against cross-origin information leakage (in [section §3.4](https://wicg.github.io/scroll-to-text-fragment/#security-and-privacy)) that appear to rule out automated, repeated pattern searches, are there still scenarios where this is an important issue? And if so, would limiting text to word boundaries really make a significant difference?

Even if both questions would be answered with yes, may it be possible and desirable to somewhat separate the concerns of expressivity and information leakage — perhaps having one algorithm that defines what fragment is being pointed at, while another defines whether this is a permissible target to scroll to? This would give flexibility and leave browsers free to not apply the limitation in situations where it can determine that there is no risk of an attack (e.g. in documents without external resources).

Are there other reasons for imposing this limitation?

### Downsides of word boundaries

I came up with at least a few reasons against this limitation, besides the obvious aspect of complicating the algorithm:

- Language specificity: As acknowledged in the spec, not every language/script has an equally clear a concept of word boundaries. Even if the `Intl.Segmenter` will be standardised and make a definition easier, it could be asked if in every language the reasoning behind the word boundary limitation is equally valid (depends on what exactly is the reasoning is).
- The limitation inhibits many valid use cases; I imagine a web-based spell checker service may want to point you at an error in a word. Or in the many documents that are not prose but other strings, there could be various reasons to point at fragments of a string, e.g. a sequence inside a chromosome.
- Browsers and/or other applications could ([should][]?) enable users to create links to selections, but the user experience for link creation would be confusing if a user can target some selections, but not others; or if alternatively the selection would be changed automagically to end at word boundaries.
[...]

Issue #99, opened 2020-03-13
[Spec] Needs to be aware of RTL


bokand:
At the very least, the spec needs to consider what should happen in BiDi scenarios. I suspect that as currently state we get some things wrong (e.g. prefix should come _logically_ before the match text).

PR #148, opened 2020-10-13
Relax word boundary requirement on prefix&suffix

Treora:
As discussed in issue #137, there is no need to require the prefix to end at a word boundary, or the suffix to start at a boundary. This PR therefore removes this requirement.

I initially intended to also refactor the algorithms as mentioned in #137: separating the word boundary check from the search, running it only after a match has been found as an ‘acceptability check’ that could then be made optional for pages without inform. However that would be quite an effort (it would require access to the concatenated text nodes again), so instead I added arguments `wordStartBounded` and `wordEndBounded` that are passed down with the appropriate values for every piece that is being searched.

Feel free to tweak any parts. In the process I made a couple of changes that may require extra attention and reconsideration:
- I tentatively turned the description of ‘locale’ into a definition because it is being used as a parameter type; but I don’t know if that is desirable. (perhaps it could better be called a ‘language tag’, while linking to ‘[language][]’ in the WHATWG HTML spec?)
- I noticed that the current spec for determining whether a substring is word bounded handles the difference between code points and code units rather confusingly. I included a commit with a suggested improvement. To simplify things it treats word boundaries as having a numerical position, much like the offset of a BoundaryPoint.
- Relatedly, I am not sure whether making a difference between a word start boundary and a word end boundary is even needed, and perhaps we could just go with a single definition instead. The existing definition tests whether the first boundary before/after the code point at _position_ is followed/preceded by the code point at _position_; couldn’t it just ask whether there is a word boundary at _position_? I did not simplify it that far as it was not clear to me why the current approach was written the way it is, and hence I might be overlooking some relevant concerns.
- I kept a definition in place for a substring being word bounded; it is however not referred to, so I suppose it could just as well be omitted. Some notes and examples could also be revisited if the concept of being word bounded is replaced by being start-bounded and end-bounded.
- I used the terms start-bounded and end-bounded, instead of left-bounded and right-bounded, to avoid assuming a specific writing direction. It sounds a bit less natural though.

I tested the changes by updating the polyfill I made, see the word-boundaries [branch][], or try it out [here][].

[language]: https://html.spec.whatwg.org/multipage/dom.html#language
[branch]: https://code.treora.com/gerben/text-fragments-ts/src/branch/word-boundaries
[here]: https://temp.treora.com/text-fragments-ts-word-boundaries/demo.html#:~:text=poi-,i,-nt

***
<a href="https://pr-preview.s3.amazonaws.com/Treora/scroll-to-text-fragment/pull/148.html" title="Last updated on Oct 22, 2020, 10:50 AM UTC (9060c8c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/scroll-to-text-fragment/148/2475249...Treora:9060c8c.html" title="Last updated on Oct 22, 2020, 10:50 AM UTC (9060c8c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -959,8 +959,9 @@ following steps:
     1. Let |potentialMatch| be null.
     1. If |parsedValues|'s [=ParsedTextDirective/prefix=] is not null:
         1. Let |prefixMatch| be the the result of running the [=find a string
-            in range=] steps given |parsedValues|'s [=ParsedTextDirective/prefix=] and
-            |searchRange|.
+            in range=] steps with |query| |parsedValues|'s
+            [=ParsedTextDirective/prefix=], |searchRange| |searchRange|,
+            |wordStartBounded| true and |wordEndBounded| false.
         1. If |prefixMatch| is null, return null.
         1. Set |searchRange|'s [=range/start=] to the first [=/boundary point=]
             [=boundary point/after=] |prefixMatch|'s [=range/start=]
@@ -979,9 +980,15 @@ following steps:
               |matchRange|'s [=range/start=] now points to the next
               non-whitespace text data following a matched prefix.
             </div>
+        1. Let |mustEndAtWordBoundary| be true if |parsedValues|'s
+            [=ParsedTextDirective/textEnd=] is non-null or
+            |parsedValues|'s [=ParsedTextDirective/suffix=] is null, false
+            otherwise.
         1. Set |potentialMatch| to the result of running the [=find a string in
-            range=] steps given |parsedValues|'s
-            [=ParsedTextDirective/textStart=] and |matchRange|.
+            range=] steps with |query| |parsedValues|'s
+            [=ParsedTextDirective/textStart=], |searchRange| |matchRange|,
+            |wordStartBounded| false, and |wordEndBounded|
+            |mustEndAtWordBoundary|.
         1. If |potentialMatch| is null, return null.
         1. If |potentialMatch|'s [=range/start=] is not |matchRange|'s
             [=range/start=], then [=iteration/continue=].
@@ -991,9 +998,15 @@ following steps:
               next instance of [=ParsedTextDirective/prefix=].
             </div>
     1. Otherwise:
+        1. Let |mustEndAtWordBoundary| be true if |parsedValues|'s
+            [=ParsedTextDirective/textEnd=] is non-null or
+            |parsedValues|'s [=ParsedTextDirective/suffix=] is null, false
+            otherwise.
         1. Set |potentialMatch| to the result of running the [=find a string in
-            range=] steps given |parsedValues|'s
-            [=ParsedTextDirective/textStart=] and |searchRange|.
+            range=] steps with |query| |parsedValues|'s
+            [=ParsedTextDirective/textStart=], |searchRange| |searchRange|,
+            |wordStartBounded| true, and |wordEndBounded|
+            |mustEndAtWordBoundary|.
         1. If |potentialMatch| is null, return null.
         1. Set |searchRange|'s [=range/start=] to the first [=/boundary point=]
             [=boundary point/after=] |potentialMatch|'s [=range/start=]
@@ -1002,9 +1015,13 @@ following steps:
         1. Let |textEndRange| be a [=range=] whose [=range/start=] is
             |potentialMatch|'s [=range/end=] and whose [=range/end=] is
             |searchRange|'s [=range/end=].
+        1. Let |mustEndAtWordBoundary| be true if |parsedValues|'s
+            [=ParsedTextDirective/suffix=] is null, false otherwise.
         1. Let |textEndMatch| be the result of running the [=find a string
-            in range=] steps given |parsedValues|'s
-            [=ParsedTextDirective/textEnd=] and |textEndRange|.
+            in range=] steps with |query| |parsedValues|'s
+            [=ParsedTextDirective/textEnd=], |searchRange| |textEndRange|,
+            |wordStartBounded| true, and |wordEndBounded|
+            |mustEndAtWordBoundary|.
         1. If |textEndMatch| is null then return null.
         1. Set |potentialMatch|'s [=range/end=] to |textEndMatch|'s
             [=range/end=].
@@ -1018,8 +1035,9 @@ following steps:
     1. Advance |suffixRange|'s [=range/start=] to the [=next non-whitespace
         position=].
     1. Let |suffixMatch| be result of running the [=find a string in range=]
-        steps given |parsedValues|'s [=ParsedTextDirective/suffix=] and
-        |suffixRange|.
+        steps with |query| |parsedValues|'s [=ParsedTextDirective/suffix=],
+        |searchRange| |suffixRange|, |wordStartBounded| false, and
+        |wordEndBounded| true.
     1. If |suffixMatch| is null then return null.
         <div class="note">
           If the suffix doesn't appear in the remaining text of the document,
@@ -1060,16 +1078,17 @@ non-whitespace position</dfn> follow the steps:
         1. Add 1 to |range|'s [=range/start offset=].
     1. If |range|'s [=range/start offset=] is equal to |node|'s
         [=Node/length=], set |range|'s [=range/start node=] to the next node in
-        [=shadow-including tree order=], and set its [=range/start offset=] to
-        0.
+        [=shadow-including tree order=], and set its [=range/start offset=] to 0.
 
-To <dfn>find a string in range</dfn> for a <a spec=infra>string</a> |query|
-in a given [=range=] |searchRange|, run these steps:
+To <dfn>find a string in range</dfn> given a <a spec=infra>string</a> |query|, a
+[=range=] |searchRange|, and booleans |wordStartBounded| and |wordEndBounded|,
+run these steps:
 
 <div class="note">
-  This algorithm will return a [=range=] that represents the first [=word
-  bounded=] instance of the |query| text that is fully contained within
-  |searchRange|. Returns null if none is found.
+  This algorithm will return a [=range=] that represents the first instance of
+  the |query| text that is fully contained within |searchRange|, optionally
+  restricting itself to matches that start and/or end at word boundaries (see
+  [[#word-boundaries]]). Returns null if none is found.
 </div>
 
 <div class="note">
@@ -1124,8 +1143,8 @@ in a given [=range=] |searchRange|, run these steps:
             |textNodeList|.
         1. Set |curNode| to the next node in [=shadow-including tree order=].
     1. Run the [=find a range from a node list=] steps given |query|,
-        |searchRange|, and |textNodeList|, as input. If the resulting [=range=]
-        is not null, then return it.
+        |searchRange|, |textNodeList|, |wordStartBounded| and |wordEndBounded|
+        as input. If the resulting [=range=] is not null, then return it.
     1. If |curNode| is null, then [=iteration/break=].
     1. [=/Assert=]: |curNode| [=tree/following|follows=] |searchRange|'s
         [=range/start node=].
@@ -1164,16 +1183,17 @@ To find the <dfn>nearest block ancestor</dfn> of a |node| follow the steps:
 1. Return |node|'s [=Node/node document=]'s [=document element=].
 
 To <dfn>find a range from a node list</dfn> given a search string |queryString|,
-a [=range=] |searchRange|, and a [=/list=] of {{Text}} nodes |nodes|, follow the
-steps:
+a [=range=] |searchRange|, a [=/list=] of {{Text}} nodes |nodes|, and booleans
+|wordStartBounded| and |wordEndBounded|, follow these steps:
 
 <div class="note">
-  This will only return a match if the matched text falls on word boundaries.
-  That is, the text match must begin and end on a word boundary. For example:
+  Optionally, this will only return a match if the matched text begins and/or
+  ends on a word boundary. For example:
 
   <div class="example">
-    “range” will match in “mountain range” but not in “color orange” nor
-    “forest ranger”.
+    The query string “range” will always match in “mountain range”, but
+    1. When requiring a word boundary at the beginning, it will not match in “color orange”.
+    2. When requiring a word boundary at the end, it will not match in “forest ranger”.
   </div>
 
   See [[#word-boundaries]] for details and more examples.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1311,20 +1327,37 @@ string. An empty string indicates that the primary language is unknown.
   determine what a valid word in the given locale is.
 </div>
 
-1. Using locale |startLocale|, let |left bound| be the last word boundary in
-    |text| that precedes |startPosition|th [=code point=] of |text|.
+To determine if an integer |position| in a <a spec=infra>string</a> |text| is a
+<dfn>word start boundary</dfn>, given a [=locale=] |locale|, follow these steps:
+
+<div class="note">
```
bokand, line 1354: Would be nice to have an intuitive description here.

bokand:
> I tentatively turned the description of ‘locale’ into a definition because it is being used as a parameter type; but I don’t know if that is desirable.

This lgtm

> I noticed that the current spec for determining whether a substring is word bounded handles the difference between code points and code units rather confusingly. I included a commit with a suggested improvement. To simplify things it treats word boundaries as having a numerical position, much like the offset of a BoundaryPoint.

Agree this was muddled, string is defined in Infra as being made up of code units so the previous text was definitely wrong. OTOH, it's not clear to me the numerical position is ever defined for boundaries? UAX29 defines them as being between words but if a boundary appears at `position`, does that mean before or after? Should we define that?

Or should we say something like (for start bound) "if `position` is the first code unit that follows `bound`"?

> couldn’t it just ask whether there is a word boundary at position?

This is related to my point above, don't you need to know in which direction? In the text `foo bar`, you wouldn't want "b" to be considered an end bound I think?

> I kept a definition in place for a substring being word bounded; it is however not referred to, so I suppose it could just as well be omitted. Some notes and examples could also be revisited if the concept of being word bounded is replaced by being start-bounded and end-bounded.

I added some non-normative notes, one of which uses this to help understand the restriction, WDYT?

> I used the terms start-bounded and end-bounded, instead of left-bounded and right-bounded, to avoid assuming a specific writing direction. It sounds a bit less natural though.

+1. I've still got #99 to actually make sure this works for different directions.

bokand (commented):
Left some comments but added a commit that addresses most, WDYT?

bokand:
Oh, and thanks for the PR - this looks great!

Treora:
Thanks for the review!

> … OTOH, it's not clear to me the numerical position is ever defined for boundaries? UAX29 defines them as being between words but if a boundary appears at `position`, does that mean before or after? Should we define that?

Inspired by [BoundaryPoint][] offsets used by Range, I would define a word boundary at position N as being right before the code unit at position N, while the position of the word boundary at the end of the string equals the string length. This also corresponds with the index a regex for `/\b/` would give you, as well as with the `index` and `index+segment.length` that the mentioned tc39 `Intl.Segmenter` [proposal][] would return.

In any case, this should be defined more accurately. I generally find the referenced [UAX29][] a rather vague spec to rely on; while implementing my polyfill I felt that more precise instructions would have been helpful.

> Or should we say something like (for start bound) "if `position` is the first code unit that follows `bound`"?

Something like that could work. It may be more wordy for the end bound, as there is no code unit after the boundary at the end of the string. But as with the current spec, such a formulation makes me feel like we are talking about some phantom object, called a ’boundary’, whose existence is only implied but which cannot be pointed at. The concept of position now seems hidden in the words “follow” and “precede”. An implementer would likely need to turn this into concrete numbers anyhow, so my suggestion is to define the concept of boundaries’ numerical position instead.

> > couldn’t it just ask whether there is a word boundary at position?
> 
> This is related to my point above, don't you need to know in which direction? In the text `foo bar`, you wouldn't want "b" to be considered an end bound I think?

I don’t know if it might be by oversight, but the current spec does not care about direction either. It talks about “a boundary preceding/following” some code point/unit, but not whether that must be a word start or word end boundary: it would just as happily let you select “Lorem ipsum ” including the space at the end (easily [verified][] in my playground).

The UAX29 spec (nor a regex) does not make differentiate between starting bounds and ending bounds, and while it is perhaps unexpected, I don’t see how a data leakage issue would come from this either. If you would want to get the more intuitive behaviour, it would require more effort in spec’ing it out. If it is okay to keep the behaviour as is, then I think that asking if there is a boundary “at *position*” would simplify the current text (if we would define numerical positions).

> I added some non-normative notes, one of which uses this to help understand the restriction, WDYT?

Good idea. I will comment in-line.


[BoundaryPoint]: https://dom.spec.whatwg.org/#boundary-points
[proposal]: https://github.com/tc39/proposal-intl-segmenter
[UAX29]: https://www.unicode.org/reports/tr29/tr29-37.html
[verified]: https://temp.treora.com/text-fragments-ts/demo.html#:~:text=Lorem%20ipsum%20

index.bs
```diff
@@ -1311,20 +1327,37 @@ string. An empty string indicates that the primary language is unknown.
...
+  |text| is assumed to be non-empty, and |position| to be a non-negative integer
+  less than or equal to its length.
```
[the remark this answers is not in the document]
Treora: Like with [BoundaryPoint][] offsets used by Range, the position would equal the string length for the word boundary at the very end of the string.

Note that the end of a string is *always* a boundary point. The current spec version already [notes][] this: “A string will always contain at least 2 word boundaries: before the first code point and after the last code point of the string.” This note seems to forget about the edge case of an empty string (hence my addition “text is assumed to be non-empty”), but otherwise this is in agreement with the first rule in the referenced word boundary definition ([UAX29][]):

> Break at the start and end of text, unless the text is empty.
> WB1 | sot | ÷ | Any
> WB2 | Any | ÷ | eot

[BoundaryPoint]: https://dom.spec.whatwg.org/#boundary-points
[notes]: https://github.com/WICG/scroll-to-text-fragment/blob/2475249f76ba20a6e6272e91853001155fdc3d22/index.bs#L1316-L1319
[UAX29]: https://www.unicode.org/reports/tr29/tr29-37.html#Word_Boundary_Rules

index.bs
```diff
@@ -1327,6 +1327,27 @@ position after its last character is a [=word end boundary=] given |endLocale|.
...
+
+  <p>
+    The goal is that a third-party must already know the full tokens they are
+    matching against. A range match like <code>textStart,textEnd</code> must be
+    word bounded on the inside of the two terms; otherwise a third party could
+    use this repeatedly to try and reveal a token (e.g. on a page with
+    <code>"Balance: 123,456 $"</code>, a third-party could set
+    <code>prefix="Balance: ", textEnd="$"</code> and vary <code>textStart</code>
+    to try and guess the numeric token.
+  </p>
```
Treora, line 1348: I think this could use a hint as to why matching could reveal anything to a third party. It could link to the possibility of information leakage in the security section?

Also, perhaps just adding “one digit at the time” at the end would help clarify that guessing the numeric token is not made impossible but only made harder. Or in even more detail: “In this example of a 6-digit number, it would be possible to find the number within 10*6 = 60 steps if there would be no word boundary requirement; while with the requirement it requires a brute-force of up to 10^6 = 1,000,000 guesses)”

index.bs
```diff
@@ -1335,8 +1356,12 @@ To determine if an integer |position| in a <a spec=infra>string</a> |text| is a
   less than or equal to its length.
 </div>
 
-1. Using |locale|, let |bound| be the last word boundary in |text| either before
-    or at |position|.
+<div class="note">
+  |position| is a word start boundary if it is the first code unit of a word.
+</div>
```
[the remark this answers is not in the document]
bokand: Ack, removed.

index.bs
```diff
@@ -1327,6 +1327,27 @@ position after its last character is a [=word end boundary=] given |endLocale|.
...
 
+<div class="example">
+  <p>
+    Text fragments are restricted such that match terms, when combined with
+    their adjacent context terms, must be word bounded. For example, in a range
+    search like <code>prefix,textStart,textEnd,suffix</code>, both
+    <code>"prefix+textStart"</code> and <code>"textEnd+suffix"</code> must be
+    word bounded. In an exact search like <code>prefix,textStart,suffix</code>,
+    <code>"prefix+textStart+suffix"</code> must be word bounded.
+  </p>
```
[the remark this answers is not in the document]
bokand: Done.

index.bs
bokand: Done. Added "one digit at a time" and linked to the doc that describes this in more detail.

bokand:
> I would define a word boundary at position N as being right before the code unit at position N

+1 - I added a sentence that says as much which hopefully makes that clear.

>>> couldn’t it just ask whether there is a word boundary at position?
>>
>> This is related to my point above, don't you need to know in which direction? In the text foo bar, you wouldn't want "b" to be considered an end bound I think?
>
>I don’t know if it might be by oversight, but the current spec does not care about direction either. It talks about “a boundary preceding/following” some code point/unit, but not whether that must be a word start or word end boundary: it would just as happily let you select “Lorem ipsum ” including the space at the end (easily verified in my playground).

Right - I guess my concern was that the boundary occurs on the correct side of the text. So in my example, you wouldn't want to allow selecting "foo b" even though the index containing 'b' has a word bound. But I think with the "boundary at position N as being right before the code unit at position N" you'd be looking for a boundary at `'b's index + 1` so this works.

On whitespace, that wasn't an intentional choice so much as whitespace isn't considered interesting so the fact that you can select it arbitrarily seems fine. I think the current behavior is fine unless there's some factor I'm missing.

index.bs
```diff
@@ -1311,20 +1327,66 @@ string. An empty string indicates that the primary language is unknown.
...
+To determine if an integer |position| in a <a spec=infra>string</a> |text| is a
+<dfn>word end boundary</dfn>, given a [=locale=] |locale|, follow these steps:
+
+<div class="note">
+  |text| is assumed to be non-empty, and |position| to be a non-negative integer
+  less than or equal to its length.
+</div>
+
+1. Using |locale|, let |bound| be the first [=word boundary=] in |text| either
+    at or after |position|.
```
bokand, line 1384: Given the definition of a boundary at position N meaning it comes before code unit at position N - should this be strictly *after* |position|? Otherwise, the 'b' in "foo bar" would be incorrectly considered a word end boundary.

Treora: Further up in the spec where it passed “*matchIndex* + *queryString*’s length - 1” as the *position*, I [changed][]  it to omit the “- 1” (this seemed simpler and may also avoid code unit/point confusion). Therefore the ‘after’ had to become ‘at/after’. So if the potential match in “foo bar” would be “b”, it would run these steps with *position* being 5 and correctly find it is not a word end boundary.

In case you meant again that e.g. “foo ” would incorrectly be considered to have a word end boundary in “foo bar”: that is the current behaviour too, as I [explained][] previously:

> > In the text foo bar, you wouldn't want "b" to be considered an end bound I think?

> I don’t know if it might be by oversight, but the current spec does not care about direction either. It talks about “a boundary preceding/following” some code point/unit, but not whether that must be a word start or word end boundary: it would just as happily let you select “Lorem ipsum ” including the space at the end (easily verified in my playground).

> The UAX29 spec (nor a regex) does not make differentiate between starting bounds and ending bounds, and while it is perhaps unexpected, I don’t see how a data leakage issue would come from this either. If you would want to get the more intuitive behaviour, it would require more effort in spec’ing it out. If it is okay to keep the behaviour as is, then I think that asking if there is a boundary “at position” would simplify the current text (if we would define numerical positions).

[changed]: https://github.com/WICG/scroll-to-text-fragment/pull/148/commits/24ddbbb14001e13235ba9586c7d69857e2b0b9d6#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985L1231-R1232
[explained]: https://github.com/WICG/scroll-to-text-fragment/pull/148#issuecomment-710692347

index.bs
```diff
@@ -1344,10 +1345,17 @@ position after its last character is a [=word end boundary=] given |endLocale|.
...
+  </p>
+  
+  <p>
+    For more details, refer to the [Security Review Doc](https://docs.google.com/document/d/1YHcl1-vE_ZnZ0kL2almeikAj2gkwCq8_5xwIae7PVik/edit#heading=h.78iny7nejmx2)
   </p>
 </div>
 
+A [=word boundary=] is at position N if it immediately precedes the code unit at
+position N.
+
```
Treora, line 1358: Great to add a definition. I suppose it should have a few words still for the case where N equals the string length.

Treora (commented):
Great. With a definition of boundary’s position, and if it is acceptable to keep the current behaviour which does not actually distinguish between word end boundaries and word start boundaries, I would simply merge the two algorithms to a single test for whether there is a boundary at position N. I will tentatively add a commit for this, which could be removed/reverted in case you’d rather change the behaviour to start making this distinction (which seems possible but more effort than I’d be willing to put into it, as the UAX29 spec doesn’t make a distinction either).

index.bs
```diff
@@ -1311,8 +1311,13 @@ language is unknown.
...
 given [=locales=] |startLocale| and |endLocale|, if both the position of its
-first character is a [=word start boundary=] given |startLocale|, and the
-position after its last character is a [=word end boundary=] given |endLocale|.
+first character [=is at a word boundary=] given |startLocale|, and the position
+after its last character [=is at a word boundary=] given |endLocale|.
+
+A number |position| <dfn>is at a word boundary</dfn> in a <a spec=infra>string</a>
+|text|, given a [=locale=] |locale|, if, using |locale|, either a [=word
+boundary=] immediately precedes the |position|th code unit, or |text|'s length
+equals |position| and |text| ends with a [=word boundary=].
```
bokand, line 1320: just a nit: "...and |text| ends with a word boundary" is describing the situation when "text's length equals position" rather than an additional condition, right? If so, it'd be clearer to remove that and instead add a sentence that the start and end of |text| are considered to be word boundaries. (and also add back that |text| must be non-empty to define a word boundary)

bokand (commented):
Thanks, I like removing the start/end distinction and the latest revision lgtm. Just a small nit in wording.

index.bs
Treora: Agreed this is unclear. I opted for “*text* ends with a word boundary” in order to leave it up to the word boundary rules to tell that a non-empty string always ends with a boundary. We do rely on that same rule ([WB1][]) to define there is a word boundary at the start of the string, so it felt a bit inconsistent to hard-code the rule for the end of the string. But your suggestion to just hard-code both start and end seems much clearer (and might be handy if a different locale somehow omits this rule).

As for requiring *text* to be a non-empty string, this should not be necessary (and might avoid the need for assertions on the side of a caller) if the definition will just return false in that case. My suggestion (I’ll commit directly):  `…or |text|'s length is more than 0 and |position| equals either 0 or |text|'s length.`

[WB1]: https://www.unicode.org/reports/tr29/tr29-37.html#WB1

Treora: If these are too many ifs, ands and ors in one sentence, perhaps it could be turned into a list of two conditions. But using the interpunction (or common sense), the sentence should be unambiguous.

bokand:
Thanks, lgtm!

PR #156, opened 2020-11-25
Add section on bidirectional behavior

### LINKED ISSUE #99: [Spec] Needs to be aware of RTL

bokand:
Describes the behavior in cases where text is non-LTR. The behavior is non-obvious but correct as specified so this just adds a non-normative section with an example to make this clearer for readers.

Also fixes mis-matched Monkeymatched directives that meant to target DOM rather than HTML specs.

Also clarifies that a few strings are expected to be ASCII strings.

Fixes #99

***
<a href="https://pr-preview.s3.amazonaws.com/bokand/ScrollToTextFragment/pull/156.html" title="Last updated on Nov 25, 2020, 2:21 AM UTC (21e3748)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/scroll-to-text-fragment/156/8bfc9f8...bokand:21e3748.html" title="Last updated on Nov 25, 2020, 2:21 AM UTC (21e3748)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -170,6 +170,59 @@ to "an example" in "this is an example text fragment", but not match to "an
 example" in "here is an example text".
 </div>
 
+### BiDi Considerations ### {#bidi-considerations}
+
+<div class='note'>This section is non-normative</div>
+
+<div class='note'>
+  See <a
+  href="https://www.w3.org/International/articles/inline-bidi-markup/uba-basics.en">Unicode
+  Bidirectional Algorithm basics</a> for a good overview of how Bidirectional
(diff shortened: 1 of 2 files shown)


bokand:
@nickburris PTAL

nickburris (approved):
LGTM

bokand:
Thanks for the quick review!

PR #158, opened 2020-11-26
Add extensions explainer with suggested selector approach.

flackr:
Explain how we might select rich content such as images and videos
through the use of a selector syntax.

### DIFF
--- a/EXTENSIONS.md
+++ b/EXTENSIONS.md
@@ -3,14 +3,14 @@
 ## Introduction
 
 The existing [scroll-to-text-fragment
-spec](https://github.com/flackr/scroll-to-text-fragment.git) enables links to
+spec](https://wicg.github.io/scroll-to-text-fragment/) enables links to
 specific textual content within a page, however there are many kinds of
 non-textual content which may also be of interest. This document explores
 several use cases and proposes methods by which they may be addressed.
 
 ## Use cases
 
-There are several content types users may be trying to view When following a
+There are several content types users may be trying to view when following a
 link to see some particular content. Primarily, these are:
 
 * Text
@@ -20,6 +20,13 @@ link to see some particular content. Primarily, these are:
 In addition to the [use cases presented for text
 content](README.md#motivating-use-cases), there are many use cases where the
 content of interest is images, video, or some element on the page.
+
+### Image aggregation or attribution
+
+Images are often collected from other sites with attribution (e.g. Wikipedia
+articles, Pinterest, Edge collections) and link back to the original content
+page. Having the ability to scroll to the image would greatly decrease the
+friction in finding that image in its original context.
 
 ### Image search engines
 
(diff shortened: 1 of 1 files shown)


EXTENSIONS.md
```diff
@@ -0,0 +1,82 @@
+# Alternate content types
+
+## Introduction
+
+The existing [scroll-to-text-fragment
+spec](https://github.com/flackr/scroll-to-text-fragment.git) enables links to
```
bokand, line 6: Should this link to the spec? i.e. https://wicg.github.io/scroll-to-text-fragment/ ?

EXTENSIONS.md
```diff
@@ -0,0 +1,82 @@
...
+specific textual content within a page, however there are many kinds of
+non-textual content which may also be of interest. This document explores
+several use cases and proposes methods by which they may be addressed.
+
+## Use cases
+
+There are several content types users may be trying to view When following a
```
bokand, line 13: nit "When" -> "when"

EXTENSIONS.md
```diff
@@ -0,0 +1,82 @@
...
+* Images
+* Videos
+
+In addition to the [use cases presented for text
+content](README.md#motivating-use-cases), there are many use cases where the
+content of interest is images, video, or some element on the page.
+
+### Image search engines
+
+Image search engines often provide the ability to view the image in the context
```
bokand, line 26: Might be worth mentioning aggregators (is there a better term for it?) like Pinterest or Edge's "Collections" feature, so we're not over-indexing (pun intended) on the Search use case.

EXTENSIONS.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Proposed solution
+
+Allow specifying a limited form of selector as part of the URL fragment using
+the same delimiter established by scroll-to-text:
+
+https://example.com#:~:selector=img[url="example.png"]
+
+Navigating to the above URL would be roughly equivalent to inserting the
+following script:
```
bokand, line 49: I think this is meant to be illustrative, yes? I expect we'd want this to set the [Indicated part of the document](https://html.spec.whatwg.org/multipage/browsing-the-web.html#scroll-to-fragid:update-the-session-history-with-the-new-page:~:text=When%20the%20user%20agent%20is%20required,is%20brought%20to%20the%20user's%20attention) as scroll-to-text and regular fragments do (thus also setting the `:target` pseudo) and following the rest of the spec.

EXTENSIONS.md
```diff
@@ -0,0 +1,82 @@
...
+## Extensions and alternatives
+
+### Video timestamps
+
+When linking to video sources it may be desirable to specify a timestamp to seek
+to. Some video services provide this capability by parsing a parameter in the
+URL, but for arbitrary video sites we could allow adding a seek parameter to
+seek to a given timestamp in the video. For example:
+
+https://example.com#:~:selector=video[url="movie.mp4"],seek=123
```
bokand, line 69: It might be good to tie this explicitly to [media fragments](https://www.w3.org/TR/media-frags/#media-fragment-syntax) so that we don't have to specify something new for the same purpose.

There is a conflict between the fragment directive using `&` to separate multiple directives, vs media fragment syntax using it to separate name-value pairs. We'd have to somehow resolve that.

bokand (approved):
Thanks, suggest some changes but lgtm otherwise

EXTENSIONS.md
flackr: Oops, yes, thanks!

EXTENSIONS.md
flackr: Done

EXTENSIONS.md
flackr: Done, thanks! I also noticed wikipedia images often have an attribution page link so this seems to fit the use case as well.

EXTENSIONS.md
flackr: Yes, this was simply meant to be illustrative. I switched the language to be consistent with the scroll-to-text proposal

EXTENSIONS.md
flackr: Done.

PR #159, opened 2020-12-02
`src` instead of `url`, correct Media Fragment use

tomayac:
- Use `src` instead of `url`, since this is what HTML uses for `` and `<video src>` (any of the URLs in the `<source>`s would match this selector, probably need to say this).
- Correct Media Fragment use (it's directly on the URL).


flackr (approved):
LGTM, thanks for the fixes

PR #165, opened 2021-02-09
Update EXTENSIONS.md

bokand:
Merge in updated proposal for CssSelector fragment

### DIFF
--- a/EXTENSIONS.md
+++ b/EXTENSIONS.md
@@ -99,10 +99,12 @@ https://example.org#:~:selector(type=CssSelector,value=img[src$="example.org"])
 The [Selectors and States as Fragment Identifiers](https://www.w3.org/TR/2017/NOTE-selectors-states-20170223/#h-frags)
 section of the above Reference Note describes the functional “selector(...)” syntax
 and [CSS Selector](https://www.w3.org/TR/2017/NOTE-selectors-states-20170223/#CssSelector_def)
-defines specifically how CSS Selectors are defined.
+defines specifically how CSS Selectors are defined. The same note also
+[describes](https://www.w3.org/TR/2017/NOTE-selectors-states-20170223/#json-examples-converted-to-fragment-identifiers)
+how to map the selectors into the fragment identifier syntax.
 
-The proposal here is to implement only type=CssSelector and start with interpreting
-only the value key.
+The proposal here is to levarage this work but implement only type=CssSelector
+and start with interpreting only the `value` key.
 
 The fragment directive allows these selectors to co-exist with pages that use the
 fragment for routing or other reasons and is already shipped in Chrome as part of
--- a/EXTENSIONS.md
+++ b/EXTENSIONS.md
@@ -115,9 +115,13 @@ text-fragments or other, future directives. E.g.
 https://example.org#:~:text=foo&selector(type=CssSelector…)&newThing
 ```
 
-The same handling as text-fragments should be used in this case. Only the first
-matching selector (from left to right in the URL) is scrolled into view (the rest
-may or may not be indicated by the UA).
+The same handling as [specified in text-fragments](https://github.com/WICG/scroll-to-text-fragment#multiple-text-directives)
+should be used in this case.
+
+_Note_: Currently, the behavior is that only the first selector (from left to
+right in the URL) is scrolled into view (the rest may or may not be indicated
+by the UA). We may wish to amend this to scroll into view the first match in
+_document order_ rather than the current _selector order_.
 
 ### CSS Selector Restrictions
 
--- a/EXTENSIONS.md
+++ b/EXTENSIONS.md
@@ -58,38 +58,148 @@ to. Here we follow the same principles as with textual content:
      additional syntax should only be used when necessary and may not be able to
      specify contrived or manufactured examples, but should extend coverage
      considerably past the most simple syntax.
+     
+## Security
+
+The issues here are analogous to those described in
+[Restrictions for scroll-to-CSS-selectors](https://docs.google.com/document/d/15HVLD6nddA0OaI8Dd0ayBP2jlGw5JpRD-njAyY1oNZo/edit#heading=h.s4z585kmzt11) and
+[Text-Fragment Security Issues](https://docs.google.com/document/d/1YHcl1-vE_ZnZ0kL2almeikAj2gkwCq8_5xwIae7PVik/edit#heading=h.uoiwg23pt0tx).
+If an attacker can navigate or convince a user to navigate to a URL with a
+“scroll-to-image” url, and if they can determine that the page scrolled automatically
+on load (or some other side effect like longer load time), they may be able to infer
+the existence of the resource on the page (with enough CSS selector syntax they could
+also infer arbitrary properties of the DOM, e.g.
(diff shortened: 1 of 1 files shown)


bokand:
cc @grantjwang

EXTENSIONS.md
```diff
@@ -58,38 +58,148 @@ to. Here we follow the same principles as with textual content:
...
+Like text-fragments, multiple such directives can be supplied, mixed with
+text-fragments or other, future directives. E.g.
+
+```
+https://example.org#:~:text=foo&selector(type=CssSelector…)&newThing
+```
+
+The same handling as text-fragments should be used in this case. Only the first
+matching selector (from left to right in the URL) is scrolled into view (the rest
+may or may not be indicated by the UA).
```
flackr, line 120: This seems debatable - might the combination of selector and text be used to refine text selection? I think there were other questions about whether anchors should be used to refine text selection.

EXTENSIONS.md
```diff
@@ -58,38 +58,148 @@ to. Here we follow the same principles as with textual content:
...
+
+### CSS Selector Restrictions
+
+The CSS selector specified in the `value=` key is restricted to a small subset of
+the selector syntax. This prevents a potential attacker from being able to reason
+about unrelated parts of a page or produce selectors with long runtimes.
+
+Selectors that do not meet the below restrictions will be blocked and the directive
+will not be invoked.
```
flackr, line 129: I wonder if we should throw an error so that the developer can understand why the scrolling hasn't happened, though I think this isn't unique to the selector restrictions as we also require a user gesture right?

EXTENSIONS.md
bokand: This is the current behavior of text fragments - it's used to highlight a logically continuous passage that has an intervening element, e.g. a run of text with an ad in-between.

I think a better way to do refinement is using the `refinedBy` property of the selector which allows nesting these selectors.

EXTENSIONS.md
bokand: We definitely could...the downside is 1) devtools already has a problem with too many errors/warnings 2) the creator and user of the fragment will likely not be the page author (or _a_ page author at all) so IMHO the value of adding an error is fairly small.

And yes, there's other restrictions like user gestures, empty browsing context group, etc.

I expect these will mostly be generated by tools so the question is will there be enough of these tools (outside Chromium developers who can use `LOG`) to make it worth it.

flackr: My thought was developers generating these links and testing them locally may be confused as to why they're not working. I agree the target site can't really do anything about the error, it's more for a person looking at dev tools.

bokand: Yeah, that makes sense - do you think we should mention it here though? It seems more like an implementation detail / something left up to UAs and not really central to the proposal.

EXTENSIONS.md
```diff
@@ -58,38 +58,148 @@ to. Here we follow the same principles as with textual content:
...
+We propose a restricted CSS selector syntax in the
+[fragment directive](https://wicg.github.io/scroll-to-text-fragment/#the-fragment-directive)
+of the URL. The selector syntax is severely restricted to avoid allowing selection
+based on arbitrary attributes or page structure.
+
+### Fragment Directive Syntax
+
+Use a slightly adapted (to fragment directives) syntax from the W3C Selectors and
+States Reference Note for the WebAnnotations [CssSelector](https://www.w3.org/TR/2017/NOTE-selectors-states-20170223/#FragmentSelector_frag).
+Here’s an example:
```
flackr, line 93: It seems to me that if this was consistent, the text fragment selector should be https://example.org#:~:selector(type=TextQuoteSelector,exact="matching text"). Is there a reason we decided the text fragment selector should not use this more verbose specification, and does that not apply here?

The proposed syntax is also somewhat removed from WebAnnotations in that we have replaced the JSON with attr=value. Given that more complex examples begin to be nested, perhaps we should use the selector JSON structure exactly (or a URI encoded version) so that developers don't need to recursively convert the structure? i.e. something like this

https://example.org#:~:selector({type:"CssSelector",value:"img[src$=\"example.org\"]"})

EXTENSIONS.md
flackr: That's fair, but if it's not a refinement, what is the intended behavior when we have multiple selectors? Should the first one in the document be selected (as with multiple matches to the text fragment)?

EXTENSIONS.md
bokand: > It seems to me that if this was consistent, the text fragment selector should be https://example.org#:~:selector(type=TextQuoteSelector,exact="matching text").

Yep - I note the consistency drawback below in the FAQ. For text fragments, we argued that simplicity and compactness were more important since they were more likely to be hand-generated. I think that's less true of a CSS selector based approach. That said, we can _also_ support TextQuoteSelector trivially.

> The proposed syntax is also somewhat removed from WebAnnotations in that we have replaced the JSON with attr=value

I'm not sure what you mean here? WebAnnotations [defines](https://www.w3.org/TR/2017/NOTE-selectors-states-20170223/#json-examples-converted-to-fragment-identifiers) a mapping of the JSON selectors to a URL fragment which we're using here so we're not really deviating by using this approach.

>  Given that more complex examples begin to be nested, perhaps we should use the selector JSON structure exactly (or a URI encoded version) so that developers don't need to recursively convert the structure?

This doesn't seem like a big hurdle to me and we'd be inventing yet another syntax which IMHO is a bigger downside.   Additionally, JSON is just one possible representation of the selector model - e.g. see [Turtle examples](https://www.w3.org/TR/2017/NOTE-selectors-states-20170223/#exturtle)

EXTENSIONS.md
bokand: For text fragments, we left the spec text with some UA defined behavior here. Only the first matched selector would be scrolled into view. In Chrome we highlight all selectors.

One possible behavior could be a find-in-page like UX where the user can cycle through all the matches. FWIW, I don't think this will be very common or useful in the non-text case so I don't think we should worry about it too much other than making sure we consider and spec the interaction between multiple directives. It's possible that in the future there could be new directives (I heard the [app-history API](https://github.com/WICG/app-history/blob/main/README.md) was considering this) and we just want to ensure this doesn't break in the presence of multiple directives.

bokand: > Should the first one in the document be selected (as with multiple matches to the text fragment)?

Ah, I misunderstood. Yeah, using the first match in document order sounds like it might be better. I think in this document we should just mention the behavior should be the same as text-fragments and change the behavior there. I did make a note of the discrepancy here though. WDYT?

EXTENSIONS.md
flackr: That seems reasonable.

flackr (approved):

EXTENSIONS.md
flackr: Sounds good!

EXTENSIONS.md
bokand: > Ah apologies, would you mind adding this link as well for context / explanation of the proposed syntax?

No worries! And done!

bokand:
Thanks!

PR #167, opened 2021-03-01
Update the font for better readability

pepelsbey:
I know it’s just an explainer, but the syntax is complicated enough on its own to further complicate it with hand-written font.

fixed

### DIFF
--- a/EXTENSIONS.md
+++ b/EXTENSIONS.md
@@ -102,7 +102,7 @@ Here is an example:
 https://example.org#:~:selector(type=CssSelector,value=img[src$="example.org"])
 ```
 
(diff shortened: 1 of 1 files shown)


pepelsbey:
@tomayac what do you think?

tomayac (approved):
LGTM, approval 

Thanks.

tomayac (changes requested):
Sorry, fired too quickly. 

The first image url should be `hills1.webp`, the second image url should be `hills2.webp`, and the selector should then select `img[src="hills2.webp"]`.

pepelsbey:
I’m sorry this font is really hard to read 😅 Will fix it right away

pepelsbey:
Fixed!

tomayac (approved):
LGTM for good now. 

Double thanks!

PR #171, opened 2021-03-22
Update outdated ID link

### LINKED ISSUE #170: Unclear link in the README
It looks like the target does not exist (anymore?), not sure what the example here is trying to illustrate. 

https://github.com/WICG/scroll-to-text-fragment#target:~:text=e.g.-,https://en.wikipedia.org/wiki/Cat%23Anatomy

tomayac:
Fixes #170

### DIFF
--- a/README.md
+++ b/README.md
@@ -477,7 +477,7 @@ details.
 ### :target
 
 For element-id based fragments (e.g.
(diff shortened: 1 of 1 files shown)


bokand:
Thanks!

PR #175, opened 2021-10-29
Explainer for FragmentDirective API

bokand:


bokand:
Also +@tomayac - would value your thoughts if you have some time to look this over.

fragment-directive-api.md
```diff
@@ -0,0 +1,211 @@
...
+
+Providing a structured API allows the browser to expose enough information and functionality to enable authors to extend and customize how different directives behave without violating either of the above goals.
+
+<sup>_1. As noted in https://crbug.com/1096983, this is accidentally exposed via the performance API. This is a bug that we'd like to fix but some use cases are currently relying on this._</sup>
+
+## Use cases
+
+* Attach comments/responses to specific parts of text on a page or provide helpful UI - e.g. [Marginalia](https://indieweb.org/marginalia)
+
+* Enable pages to easily create text directive links. The rules for how text is matched are [necessarily complicated](https://wicg.github.io/scroll-to-text-fragment/#find-a-range-from-a-text-directive); they must consider word boundaries, DOM node display types and visibility and various nuances of how DOM is traversed. This API allows an author let the browser generate a valid text directive URL for a given Range.
```
tomayac, line 25:
```suggestion
* Enable pages to easily create text directive links. The rules for how text is matched are [necessarily complicated](https://wicg.github.io/scroll-to-text-fragment/#find-a-range-from-a-text-directive); they must consider word boundaries, DOM node display types and visibility, and various nuances of how DOM is traversed. This API allows an author to let the browser generate a valid text directive URL for a given Range.
```

fragment-directive-api.md
```diff
@@ -0,0 +1,211 @@
...
+
+* Enable text directives in cross-origin iframes. To prevent [XS-Search attacks](https://wicg.github.io/scroll-to-text-fragment/#example-4d0b486d:~:text=A%20malicious%20page%20embeds%20a%20cross%2Dorigin%20victim%20in%20an%20iframe,its%20own%20document.), text directives are not applied when navigated from a cross-origin initiator. However, an iframe can navigate itself to a text directive. By allowing an embedder page to read the text directive, it can postMessage it to a cross-origin document that's opted-in to this behavior, enabling deep linking to the inner frame (see examples section below).
```
tomayac, line 27:
```suggestion
* Enable text directives in cross-origin iframes. To prevent [XS-Search attacks](https://wicg.github.io/scroll-to-text-fragment/#example-4d0b486d:~:text=A%20malicious%20page%20embeds%20a%20cross%2Dorigin%20victim%20in%20an%20iframe,its%20own%20document.), text directives are not applied when navigated from a cross-origin initiator. However, an iframe can navigate itself to a text directive. By allowing an embedder page to read the text directive, it can `postMessage()` it to a cross-origin document that's opted-in to this behavior, enabling deep linking to the inner frame (see examples section below).
```

fragment-directive-api.md
```diff
@@ -0,0 +1,211 @@
...
+  constructor(TextDirectiveOptions);
+  // TODO: constructor(DOMString directive_string);
+  readonly attribute DOMString prefix;
+  readonly attribute DOMString textStart;
+  readonly attribute DOMString textEnd;
+  readonly attribute DOMString suffix;
+};
+```
+
+Why a SelectorDirective base-class, in addition to Directive? The [proposed](https://github.com/WICG/scroll-to-text-fragment/blob/main/EXTENSIONS.md#proposed-solution) CSS selector directive would behave very similarly to a text directive and allows createSelectorDirective to return a SelectorDirective. OTOH, the proposed [note selector](https://github.com/bokand/web-annotations/blob/main/URL-based-annotation.md) would not fit this interface.
```
tomayac, line 79:
```suggestion
Why a `SelectorDirective` base-class, in addition to `Directive`? The [proposed](https://github.com/WICG/scroll-to-text-fragment/blob/main/EXTENSIONS.md#proposed-solution) CSS selector directive would behave very similarly to a text directive and allows `createSelectorDirective()` to return a `SelectorDirective`. OTOH, the proposed [note selector](https://github.com/bokand/web-annotations/blob/main/URL-based-annotation.md) would not fit this interface.
```

fragment-directive-api.md
```diff
@@ -0,0 +1,211 @@
...
+
+_TODO: Maybe SelectorDirective is unnecessary? Callers could always determine the directive type using `Directive.type` if they need. Also, it may actually make sense for `note` to provide `getMatchingRange`...)_
```
tomayac, line 81:
```suggestion
_TODO: Maybe `SelectorDirective` is unnecessary? Callers could always determine the directive type using `Directive.type` if they need to. Also, it may actually make sense for `note` to provide `getMatchingRange()`...)_
```

fragment-directive-api.md
```diff
@@ -0,0 +1,211 @@
...
+  };
+};
+```
+
+### Forward a text directive across origins
+
+```JS
+// Embedder document
+const text_directives =
+    document.fragmentDirectives.items.filter(i => i.type == "text"));
```
tomayac, line 118:
```suggestion
    document.fragmentDirectives.items.filter(i => i.type === "text"));
```

fragment-directive-api.md
```diff
@@ -0,0 +1,211 @@
...
+
+frames[0].postMessage(message);
+```
+
+In the cross-origin document:
+
+```JS
+//Embedee document
+window.onmessage = (e) => {
+  if (e.type == 'text-directives') {
```
tomayac, line 133:
```suggestion
  if (e.type === 'text-directives') {
```

fragment-directive-api.md
```diff
@@ -0,0 +1,211 @@
...
+    window.location.hash = `:~:${strings.join('&')}`;
+  }
+});
+```
+
+_TODO: setting `location.hash` isn't great. Consider adding `fragmentDirective.add(Directive)` and adding a `Directive.remove()`._
+
+## FragmentDirective.items
+
+The `items` array reflects the currently active directives on the page. Using text directives as an example, an entry should exist in `items` for a text directive as long a highlight is showing. If the user dismisses the highlight it is removed from the array. Conversely, if the directive is removed from `items` programmatically (see next section), the highlight should be removed from the page.
```
tomayac, line 144:
```suggestion
The `items` array reflects the currently active directives on the page. Using text directives as an example, an entry should exist in `items` for a text directive as long a highlight is showing. If the user dismisses the highlight, it is removed from the array. Conversely, if the directive is removed from `items` programmatically (see next section), the highlight should be removed from the page.
```

fragment-directive-api.md
```diff
@@ -0,0 +1,211 @@
...
+
+## FragmentDirective as part of location.hash
+
+Currently, script can add a directive by writing to `location.hash`:
+
+```JS
+location.hash = ":~:text=foo%20bar";
+```
+
+Will add a text directive to the page, highlighting "foo bar" and adding a `TextDirective` to `fragmentDirective.items`. However, this still runs the [fragment directive stripping steps](https://wicg.github.io/scroll-to-text-fragment/#process-and-consume-fragment-directive):
```
tomayac, line 154:
```suggestion
The snippet above will add a text directive to the page, highlighting "foo bar" and adding a `TextDirective` to `fragmentDirective.items`. However, this still runs the [fragment directive stripping steps](https://wicg.github.io/scroll-to-text-fragment/#process-and-consume-fragment-directive):
```

fragment-directive-api.md
```diff
@@ -0,0 +1,211 @@
...
+location.hash = value;
+console.log(location.hash);  // Output: ""
+```
+
+This is rather unintuitive and surprising.
+
+There's also the question of what happens to existing directives in `fragmentDirective.items` when the hash is modified. In the cases below, suppose the user navigated to `https://example.org/blog.html#:~:text=acme`.
+
+1. What should happen when script sets a hash with no fragment directive? (e.g. `location.hash = 'page1';`).
+2. What should happen when script sets a hash with an unrelated directive? (e.g. `location.hash = ':~:note(href=notes.example.org);'`)
```
tomayac, line 167:
```suggestion
2. What should happen when script sets a hash with an unrelated directive? (e.g. `location.hash = ':~:note(href=notes.example.org)';`)
```

fragment-directive-api.md
```diff
@@ -0,0 +1,211 @@
...
+3. What should happen when script sets a hash with a text directive? (e.g. `location.hash = ':~:text=blog%20title';`)
+
+That is, are changes to `location.hash` additive or do they replace existing directives?
+
+For case 1, we almost certainly shouldn't affect existing directives as this would violate the _compatibility_ goal from the introduction. Pages often write to their hash for various reasons, these shouldn't interfere with user-supplied directives. That is, a page modifying its hash shouldn't remove text highlights.
+
+For case 2, it also seems like we shouldn't remove the text directive. Directives of different types should behave independently. That is, adding an annotation to a page shouldn't clear text highlights.
+
+In case 3 either behavior could work: a new highlight should be added and the existing one kept OR the new highlight replaces all existing ones. Though, if additive, it means there's no way to remove directives.
```
tomayac, line 176:
```suggestion
In case 3, either behavior could work: a new highlight should be added and the existing one kept OR the new highlight replaces all existing ones. Though, if additive, it means there's no way to remove directives.
```

tomayac (commented):
I think this goes in the right direction.

- I think the IDL needs to state where it's exposed (`document`).
- There's some redundancy of repeating the directive name `text=` when the type is known (`new TextDirective("text=foo")` vs. a hypothetical simpler `new TextDirective("se")`), but I think that's fine.

fragment-directive-api.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Current Status
+
+As of Oct 29, 2021: The API described below is available for experimentation in Chrome 97.0.4685.0 and newer behind a flag. Use `--enable-blink-features=TextFragmentAPI` to turn it on (or chrome://flags/#enable-experimental-web-platform-features which turns on all experimental features).
+
+## Introduction
+
+This document proposes a programmatic API through which authors can interact with text (and future) directives.
+
+Today, when a page is loaded with a text directive such as `https://example.org#:~:text=foo,bar`, the author has no way<sup>1</sup> to tell that a text directive was set or what text was highlighted. The fragment directive portion of the URL (everything in the fragment after and including `:~:`) is stripped from the URL when the document is loaded. This is done for two reasons:
```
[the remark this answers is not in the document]
bokand: Neat! Thanks!

bokand:
Thanks Thomas!

> I believe the IDL needs to state where it's exposed (document).

You're right, though it's `[Exposed=Window]` since `Exposed` works on a global object. FragmentDirective is already spec'ed in our draft spec as being a member of `Document` - I added the existing IDL here for clarity. Maybe it'd make sense to also expose `TextDirective` and related on `Worker` as well? `getMatchingRange` couldn't work but the attributes might be useful? Thoughts?

> There's some redundancy of repeating the directive name text= when the type is known (new TextDirective("text=foo") vs. a hypothetical simpler new TextDirective("foo")), but I think that's fine.

Good catch, I have a mix of both styles here - I agree the latter is more ergonomic so I changed the examples to that.

It might be useful to have a `Directive.parse(DOMString)` that just returns a list of `Directive` objects of the right type (e.g. so you could do `Directive.parse("text=foo&note(href=example.com)&selector(...)")` and get back 3 `Directive` objects) but I'm weary of adding too many bells and whistles before we know how it'll be used.

bokand:
@flackr please also take a look

fragment-directive-api.md
```diff
@@ -0,0 +1,230 @@
...
+
+_TODO: Maybe `SelectorDirective` is unnecessary? Callers could always determine the directive type using `Directive.type` if they need to. Also, it may actually make sense for `note` to provide `getMatchingRange()`...)_
+
+## Examples
+
+### Marginalia-like use cases:
+
+```JS
+// Coming from a server-side WebMention:
+const comment_text = "Great Point!";
```
flackr, line 108: How is this use case different from a note in which case the content would be hidden?

Or perhaps more specifically, why would the location of the note be coming from a fragmentdirective but the note content be coming from another server?

fragment-directive-api.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Use cases
+
+* Attach comments/responses to specific parts of text on a page or provide helpful UI - e.g. [Marginalia](https://indieweb.org/marginalia)
+
+* Enable pages to easily create text directive links. The rules for how text is matched are [necessarily complicated](https://wicg.github.io/scroll-to-text-fragment/#find-a-range-from-a-text-directive); they must consider word boundaries, DOM node display types and visibility, and various nuances of how DOM is traversed. This API allows an author to let the browser generate a valid text directive URL for a given Range.
+
+* Enable text directives in cross-origin iframes. To prevent [XS-Search attacks](https://wicg.github.io/scroll-to-text-fragment/#example-4d0b486d:~:text=A%20malicious%20page%20embeds%20a%20cross%2Dorigin%20victim%20in%20an%20iframe,its%20own%20document.), text directives are not applied when navigated from a cross-origin initiator. However, an iframe can navigate itself to a text directive. By allowing an embedder page to read the text directive, it can `postMessage()` it to a cross-origin document that's opted-in to this behavior, enabling deep linking to the inner frame (see examples section below).
+
```
flackr, line 28: Another use case is to provide application specific helpful UI. E.g. a sublime like editor might highlight sections of the preview containing notes, or an application could provide an arrow that points to the fact that there are notes / text to be read further down, possibly also jumping to that note when clicked.

fragment-directive-api.md
```diff
@@ -0,0 +1,230 @@
...
+// See https://wicg.github.io/scroll-to-text-fragment/#feature-detectability
+
+// === Current ===
+
+[Exposed=Window]
+interface FragmentDirective {
+};
+
+partial interface Document {
+    [SameObject] readonly attribute FragmentDirective fragmentDirective;
```
flackr, line 44: Document seems like a strange place to access the current fragmentdirective. Wouldn't these API's to access / modify the fragment directive make more sense as part of Location?

fragment-directive-api.md
bokand: This is describing a separate idea/process to annotations than we're working on - see https://indieweb.org/marginalia. 

In the indieweb world, blog A can make reference to something on blog B and notify blog B by sending it a web mention. If this is in reply to a specific part of Blog B, the webmention needs a way to specify that, i.e. specify what range of text it's replying to. Blog B's server can then go and fetch the content from blog A and attach it to the range specified in the text directive.

marginalia requires some way for blog A to make reference to a specific part of blog B. The indieweb folks envisaged using [fragmention](https://indieweb.org/fragmention) for this but text directives solve the same problem and are actually shipped. In their case, blog A has opted into sharing the reply/annotation with blog B which isn't the case (by default) for our user-generated annotations.

fragment-directive-api.md
bokand: Great idea - added

fragment-directive-api.md
bokand: Heh - indeed. Alas, this didn't work out. See crbug.com/1057795 and https://github.com/whatwg/html/issues/5523 for some context.

`window.location` has a bunch of legacy baggage and behaves in some surprising ways that can lead to leaking objects like this across origins if we're not careful; `Location` needs to refer to the active document in the browsing context rather than its own.

bokand:
@flackr Since I'll be out for the next 6 weeks I'm going to merge the latest changes so I have something to point to in the intent to prototype but feel free to make changes while I'm gone if you have any.

PR #183, opened 2022-01-24
[Alternative Content Types proposal] Allow `:has()` pseudo class

tomayac:
We found `has:()` to be very useful to select nested structures like `<video><source src="foo" /></video>` based on the `src` attribute.

https://github.com/tomayac/link-to-media/blob/62eb5f0aec182382e9a7b37682f8da5cdbda0815/polyfill/create-link.js#L50-L57


tomayac:
FYI: @samdutton @rreverser

PR #191, opened 2022-08-09
Reset start offset in 'find a string in range' steps

### LINKED ISSUE #188: Need to specify setting start offset to zero
in the description of the "To find a string in range" subroutine, we have these two parallel steps:

If curNode is part of a non-searchable subtree:
If curNode is not a visible text node:

In the first it is not explicitly stated that the start offset needs to be set to zero. It does explicitly say that in the second step. It should say that start offset needs to be set to zero in both steps.

bokand:
Fixes #188

***
<a href="https://pr-preview.s3.amazonaws.com/bokand/ScrollToTextFragment/pull/191.html" title="Last updated on Aug 9, 2022, 2:41 PM UTC (bfc7f8c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/scroll-to-text-fragment/191/c5a7035...bokand:bfc7f8c.html" title="Last updated on Aug 9, 2022, 2:41 PM UTC (bfc7f8c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1403,11 +1403,12 @@ run these steps:
             1. Set |searchRange|'s [=range/start node=] to the next node, in
                 [=shadow-including tree order=], that isn't a [=shadow-including
                 descendant=] of |curNode|.
+            1. Set |searchRange|'s [=range/start offset=] to 0.
             1. [=iteration/Continue=].
         1. If |curNode| is not a [=visible text node=]:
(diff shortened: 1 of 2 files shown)


bokand:
@megangardner - ptal?

bokand:
@megangardner friendly ping

bokand:
I went ahead and merged these since they're small but PLMK if you find an issue and I can address it later.

PR #192, opened 2022-08-09
Fix advancement in 'Next non-whitespace position'.

### LINKED ISSUE #189: Need additional continue in advancing to next non-whitespace position
After implementing the subroutine as described here:
https://wicg.github.io/scroll-to-text-fragment/#next-non-whitespace-position

I had issues when I was starting this subroutine at the end of a node, like when parsing this URL:
https://blog.chromium.org/2019/12/chrome-80-content-indexing-es-modules.html#:~:text=the-,text,-parameter

Therefore, I added a step before 1.3. that was molded after 1.3 and 1.4 which was
 - If offset is equal to node length, do the same steps 1.4.1 and 1.4.2
 
 I believe this is necessary to have the correct functionality when starting at the end of a node.

bokand:
These steps didn't consider the case where they're run on a range whose
start is already at the end of its node.

Moving the node advancement step to the start of the loop fixes this.

Also consolidate the node advancement for various cases into a single
check. Previously the non-searchable subtree case would advance by 'next
node that isn’t a shadow-including descendant of node' which is now
omitted in the generalized case. However, this is equivalent since if
the next node is a shadow-including descendant of node, it must also be
in a non-searchable subtree and so it will keep advancing until it is no
longer a descendant of node.

Fixes #189

***
<a href="https://pr-preview.s3.amazonaws.com/bokand/ScrollToTextFragment/pull/192.html" title="Last updated on Sep 2, 2022, 1:21 PM UTC (31e000e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/scroll-to-text-fragment/192/9f532e0...bokand:31e000e.html" title="Last updated on Sep 2, 2022, 1:21 PM UTC (31e000e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1331,15 +1331,10 @@ non-whitespace position</dfn> follow the steps:
     1. While |range| is not collapsed:
         1. Let |node| be |range|'s [=range/start node=].
         1. Let |offset| be |range|'s [=range/start offset=].
-        1. If |node| is part of a [=non-searchable subtree=] then:
-            1. Set |range|'s [=range/start node=] to the next node, in
-                [=shadow-including tree order=], that isn't a [=shadow-including
-                descendant=] of |node|, and set its [=range/start offset=] to 0.
-            1. [=iteration/Continue=].
-        1. If |node| is not a [=visible text node=]:
-            1. Set |range|'s [=range/start node=] to the next node, in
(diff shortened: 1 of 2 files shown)


bokand:
@megangardner friendly ping

bokand:
I went ahead and merged this since it's small but PLMK if you find an issue.

PR #201, opened 2022-12-21
[Spec] export the directive stripping steps

### LINKED ISSUE #198: Define "strip the fragment directive" generically, without the extra Document steps
I'm defining how [fragment-only CSS urls (like `url(#foo)`) work](https://w3c.github.io/csswg-drafts/css-values-4/#local-urls), and need to get the fragment so I can search for a matching ID in the tree. If the url is `#foo:~:text=foo`, presumably I want to strip the directive and just search for "foo" as the ID, but the spec doesn't currently define a generic algorithm allowing me to do this; it bundles this behavior into also updating some Document state.

Could I get that algo abstracted out, so it takes a raw fragment string and returns a pair of (fragment, directive) strings?

bokand:
Separates the algorithm for splitting off the fragment directive out of a raw fragment into a separate exported algorithm usable from other specs.

Fixes #198

***
<a href="https://pr-preview.s3.amazonaws.com/bokand/ScrollToTextFragment/pull/201.html" title="Last updated on Dec 21, 2022, 2:43 AM UTC (489e0ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/scroll-to-text-fragment/201/694193f...bokand:489e0ff.html" title="Last updated on Dec 21, 2022, 2:43 AM UTC (489e0ff)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -268,6 +268,28 @@ To the definition of [=Document=], add:
 >     resource. It is initially null.
 >   </em>
 
+<div algorithm="split the fragment from the fragment directive">
+  To <dfn export>split the fragment from the fragment directive</dfn>, given an ASCII string
+  |raw fragment| and returning a [=/tuple=] consisting of a <code>fragment</code> and a
+  <code>fragment directive</code> (both ASCII strings), run these steps:
+
+  <ol class="algorithm">
+    1. Let |position| be the [=string/position variable=] pointing to the first code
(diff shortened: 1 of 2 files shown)


bokand:
@tabatkins, does this look ok?

bokand:
I'm doing a lot of cross-cutting work on the spec so I'm going to go ahead and merge it to avoid merge headaches. But happy to followup with any feedback you may have.

PR #203, opened 2022-12-21
[Spec] Fix naming of textFragmentActivationFlag

### LINKED ISSUE #178: Consider renaming textFragmentToken
Token typically does not mean a boolean in the web platform. At least in my experience. "Capability" might be more appropriate.

bokand:
As per #178, spec text now typically uses space-separated names and "flag" is discouraged so this PR fixes #202 by changing `textFragmentActivationFlag` to `text fragment activation` and `allowTextFragmentDirective` to `allow text fragment directive` and fixes up some surrounding prose.

Fixes #178

***
<a href="https://pr-preview.s3.amazonaws.com/bokand/ScrollToTextFragment/pull/203.html" title="Last updated on Dec 22, 2022, 8:53 PM UTC (eceafd7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/scroll-to-text-fragment/203/b0837c5...bokand:eceafd7.html" title="Last updated on Dec 22, 2022, 8:53 PM UTC (eceafd7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -699,37 +699,35 @@ to find and set the [=Document=]'s indicated part.
 ### Restricting the Text Fragment ### {#restricting-the-text-fragment}
 
 Amend the definition of a [=/request=] and of a [=Document=] to include a new
-[=document/textFragmentActivationFlag=] field:
+boolean [=document/text fragment activation=] field:
 
 >   <strong>Monkeypatching [[FETCH]]:</strong>
 >
->   A [=/request=] has an associated <dfn for="request">textFragmentActivationFlag</dfn>.
+>   A [=/request=] has an associated boolean <dfn for="request">text fragment activation</dfn>,
+>   initially false.
 
 >   <strong>Monkeypatching [[HTML]]:</strong>
 >
->   A [=Document=] has a <dfn for="document">textFragmentActivationFlag</dfn> that is
->   consumed in order to allow a single activation of a text fragment. This flag is
->   set during document loading only if the navigation occurred as a result of a user
->   activation.
+>   Each [=Document=] has a <dfn for="document">text fragment activation</dfn>, which is a boolean,
+>   initially false.
 >
->   If the [=Document=]'s [=document/textFragmentActivationFlag=] isn't consumed to activate
->   a text fragment, it may be consumed to set the [=request/textFragmentActivationFlag=]
->   flag of a navigation [=/request=]. In this way, a [=document/textFragmentActivationFlag=]
->   can be propagated from one [=Document=] to another across a navigation.
+>   <div class="note">
+>     [=document/text fragment activation=] is consumed in order to allow a single activation of a
+>     text fragment. It is set during document loading only if the navigation occurred as a result
+>     of a user activation and can be propagated across client-side redirects.
 >
->   Reading either the [=Document=]'s [=document/textFragmentActivationFlag=] or the
->   [=/request=]'s [=request/textFragmentActivationFlag=] flag must always consume the value by
->   setting it to false, such that the flag cannot be reused.
+>     If a [=Document=]'s [=document/text fragment activation=] isn't consumed to activate
+>     a text fragment, it may be consumed to set the [=request/text fragment activation=]
+>     flag of a navigation [=/request=]. In this way, a [=document/text fragment activation=]
+>     can be propagated from one [=Document=] to another across a navigation.
+>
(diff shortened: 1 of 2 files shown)


bokand:
@annevk - I can't add you as a reviewer but PTAL.

index.bs
```diff
@@ -699,37 +699,35 @@ to find and set the [=Document=]'s indicated part.
...
+>   Each [=Document=] has a <dfn for="document">text fragment activation</dfn>, which is a boolean,
+>   initially false.
 >
->   If the [=Document=]'s [=document/textFragmentActivationFlag=] isn't consumed to activate
->   a text fragment, it may be consumed to set the [=request/textFragmentActivationFlag=]
->   flag of a navigation [=/request=]. In this way, a [=document/textFragmentActivationFlag=]
->   can be propagated from one [=Document=] to another across a navigation.
+>   <div class="note">
+>     [=document/text fragment activation=] is consumed in order to allow a single activation of a
+>     text fragment. It is set during document loading only if the navigation occurred as a result
```
annevk, line 716: You'll need to update set/unset prose to talk about true/false instead.

At the same time I might try to rephrase the bits about consumption. E.g.,

> In order to allow a single activation of a text fragment, documents have an associated text fragment activation. It is true ...

index.bs
```diff
@@ -762,20 +760,23 @@ Amend the definition of a [=/request=] and of a [=Document=] to include a new
 
 >   <strong>Monkeypatching [[HTML]]:</strong>
 >
->   A [=Document=] has an <dfn for="document">allowTextFragmentDirective</dfn>
->   flag that is used to determine whether a text fragment directive should be
->   allowed to activate. If this flag is false, the text fragment must not
->   cause any observable effects.
+>   Each [=Document=] has an <dfn for="document">allow text fragment</dfn>, which is a
```
annevk, line 763: maybe "allow text fragment activation" given that's its purpose? It doesn't disable text fragments entirely presumably.

annevk (commented):
Generally looks quite a bit better, thanks!

index.bs
bokand: Done.

index.bs
bokand: I'm a bit worried about confusion between `allow text fragment activation` and `text fragment activation`. It's a bit verbose but WDYT about calling this one `allow observable text fragment effects`?

annevk: I liked that they both had activation as they seemed related, or are they not? Could also call the other one "text fragment activation happened" maybe.

I also noticed that these notes contain conformance terms, which isn't allowed. Bikeshed usually warns/errors about that.

(There's also a somewhat weird suggestion that reading the field causes it to change, which isn't quite true. I guess it means that call sites carefully flip values after observing them?)

annevk:
(To give better feedback I probably need to reread the document, but that'll have to wait until after the holidays.)

index.bs
bokand: > I liked that they both had activation as they seemed related, or are they not? Could also call the other one "text fragment activation happened" maybe.

`text fragment activation` - was the navigation, potentially across client-redirects, user-activated? (maybe it'd be clearer as `text fragment user activation`?)

`allow text fragment`  - looking at all of: the navigation's initiator origin, browsing context set size, and `text fragment activation`, is it safe to allow scrolling to the text fragment? This is computed when the document loads but used later when performing "scroll to the fragment". We compute it during document loading rather than when used since we don't have access to NavigationParams when used in "scroll to the fragment". (I was trying to keep it generic but scrolling is the only effect we're currently interested in so maybe better described by `allow text fragment scrolling`?)

> I also noticed that these notes contain conformance terms, which isn't allowed. Bikeshed usually warns/errors about that.

Will rephrase. Odd, I get a clean bill from Bikeshed...

> (There's also a somewhat weird suggestion that reading the field causes it to change, which isn't quite true. I guess it means that call sites carefully flip values after observing them?)

Yeah, it just means callsites always set to false when reading. I'll clarify.

bokand:
> (To give better feedback I probably need to reread the document, but that'll have to wait until after the holidays.)

No worries - it'll hopefully be in better shape by then. Enjoy the holiday break!

bokand:
Since I'm doing lots of work here, I'm going to go ahead and merge what I've got to avoid merge headaches, but happy to take more feedback when you return (and of course, it'll go through further review when upstreaming).

Issue #221, opened 2023-03-27
Should the processing model care about validity?


annevk:
It currently has

> If fragment directive input is not a [valid fragment directive](https://wicg.github.io/scroll-to-text-fragment/#valid-fragment-directive), then return an empty [list](https://infra.spec.whatwg.org/#list).

but can that actually happen? It seems better if we don't need these additional grammar-based validation requirements.

bokand:
Hmm, are you suggesting to remove the grammar definitions and rely on the parsing and processing steps to handle errors?

That makes sense to me, though personally I'd find the grammar useful so readers can tell at a glance how to construct valid text directives (or other fragment directives); maybe keep that as a non-normative reference?

annevk:
Yeah, I think a dual nature makes sense. One to guide producers and the other to guide consumers. URL and the HTML syntax also have this dual nature.

bokand:
One additional point, I see that for `FragmentDirective` it doesn't matter since an invalid string will just be an `UnknownDirective`. But for `TextDirectives` we do use the grammar to discard invalid cases (e.g. a URL includes a fifth, invalid, term after `suffix`). That seems convenient to have a grammar? But I'll see if it's easy to just do algorithmically.

annevk:
Usually the problem is that implementers don't use grammars in their implementation so you get subtle differences. I've found a lot of these over the years in HTTP header parsers.

They also had quite a bit of complexity when a string iterator can do the job as well.

bokand:
Yeah, that's true, but then I have a bit of a meta question about spec writing :) ...I recently came across [this note in parallelism](https://html.spec.whatwg.org/multipage/infrastructure.html#parallelism:~:text=Implementations%20are%20not%20expected%20to%20implement%20this%20as%20a%20continuously%20running%20loop.%20Algorithms%20in%20standards%20are%20to%20be%20easy%20to%20understand%20and%20are%20not%20necessarily%20great%20for%20battery%20life%20or%20performance.):

```
Algorithms in standards are to be easy to understand and are not necessarily great for battery life or performance.
```

How do we balance this vs. writing steps that match implementation? Blink also doesn't use a grammar for this case but this seemed like the best way to convey the requirements in spec; implementors can validate however they choose but the spec wins in case of any deviations (does HTML follow something similar to [C++'s as-if rule](https://en.cppreference.com/w/cpp/language/as_if)?).

Is this typically a judgement call, with preference to match implementation?

[...]

PR #222, opened 2023-03-27
Disambiguate UnknownDirective grammar

bokand:
The grammar for an `UnknownDirective` accidentally allows `&` which would make it ambiguous, since directives are separated by `&`. Remove `&` from the set of allowed characters.


annevk:
I guess this illustrates that my intuition in #221 is correct.

PR #225, opened 2023-04-07
Improve integration with navigation

bokand:
This PR makes it so that the fragment directive is stripped from the URL whenever it is set to a history entry. Instead, the history entry keeps it in a separate `directive state`.

The `directive state` is shared between contiguous history entries where the only difference is the non-directive portion of the fragment.

Moving between session history entries will cause the `directive state`, if it differs from the current entry, to be set into an `uninvoked directives` member on the Document. The "scroll to the fragment" steps will read directives from this member and attempt to perform the text search.

When the fragment search ends, `uninvoked directives` is always cleared.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -59,6 +59,16 @@ spec:url; type:dfn; text:fragment
   }
 </pre>
 
+<style>
+  .monkeypatch {
+    color: grey;
+  }
+
+  .monkeypatch .diff {
+    color: black;
+  }
+</style>
+
 <h2 id=infrastructure>Infrastructure</h2>
 
 <p>This specification depends on the Infra Standard. [[!INFRA]]
@@ -237,12 +247,12 @@ The <dfn>fragment directive delimiter</dfn> is the string ":~:", that is the
 three consecutive code points U+003A (:), U+007E (~), U+003A (:).
 
 <div class="note">
-  The [=/fragment directive=] is part of the URL fragment. This means it
+  The [=fragment directive=] is part of the URL fragment. This means it
   always appears after a U+0023 (#) code point in a URL.
 </div>
 
 <div class="example">
-  To add a [=/fragment directive=] to a URL like https://example.com, a fragment
+  To add a [=fragment directive=] to a URL like https://example.com, a fragment
   is first appended to the URL: https://example.com#:~:text=foo.
 </div>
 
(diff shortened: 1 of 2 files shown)


bokand:
@jakearchibald @annevk, how does this look?

index.bs
```diff
@@ -260,159 +270,280 @@ action. Multiple directives may appear in the fragment directive.
...
+>     This algorithm makes a URL's fragment end at the [=fragment directive
+>     delimiter=]. The returned [=/fragment directive=] includes all characters that follow the
+>     delimiter but does not include the delimiter.
+>   </div>
+>
+>   <div class="issue">
+>     If a URL's fragment ends with ':~:' (i.e. empty directive), this will return null which is
+>     treated as the URL not specifying an explicit directive (and avoids clobbering an existing
+>     one. But maybe in this case we should return the empty string? That way a page can explicitly
+>     clear directives/highlights by navigating/pushState to '#:~:'.
```
jakearchibald, line 362: I agree with `#:~:` clearing current directives

index.bs
```diff
@@ -59,6 +59,16 @@ spec:url; type:dfn; text:fragment
...
 </pre>
 
+<style>
+  .monkeypatch {
+    color: grey;
+  }
+
+  .monkeypatch .diff {
+    color: black;
+  }
```
jakearchibald, line 69: This creates black text on a black background for uses with dark mode enabled.

index.bs
```diff
@@ -260,159 +270,280 @@ action. Multiple directives may appear in the fragment directive.
...
->   9. Run the [=process and consume fragment directive=] steps on
->     |creationURL| and |document|.
->   10. Set |document|'s [=Document/URL=] to be |creationURL|.
+>   <div class="monkeypatch">To navigate to a fragment given navigable |navigable|, ...:
+>     1. <span class="diff">Let |directive state| be navigable's active session history
+>         entry's [=she/directive state=].</span>
+>     1. <span class="diff">Let |fragment directive| be the result of running
+>         [=remove the fragment directive=] on |url|.</span>
+>     1. <span class="diff">If |url| does not equal |navigable|'s active session history
+>         entry's URL with exclude fragments set to true OR |fragment directive| is not null:</span>
```
jakearchibald, line 428: Aren't we guaranteed that the URLs are equal, excluding fragments? Else it isn't a fragment navigation.

jakearchibald (changes requested):
Looks good! Mostly small changes needed. And handling the redirect case.

index.bs
```diff
@@ -260,159 +270,280 @@ action. Multiple directives may appear in the fragment directive.
...
-    1. Return the [=/tuple=] (|fragment|, |fragmentDirective|).
-  </ol>
-</div>
+>   <dfn>directive state</dfn> holds the value of the [=fragment directive=] at the time the session
+>   history entry was created and is used to invoke directives, such as text highlighting, whenever
+>   the entry is traversed. It has:
+>   * <dfn for="directive state">value</dfn>, the [=fragment directive=] ASCII string or null,
+>       initially null.
+>
+>   A [=/directive state=] may be shared across contiguous session history entries.
```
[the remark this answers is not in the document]
bokand: :exploding_head:

Heh, zany. I assumed contiguous based on [the text in HTML](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-and-session-history:~:text=All%20entries%20that%20share%20the%20same%20document%20state%20(and%20that%20are%20therefore%20merely%20different%20states%20of%20one%20particular%20document)%20are%20contiguous%20by%20construction.), we should probably update that?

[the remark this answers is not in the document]
jakearchibald: Hah, good catch. Not sure how that snuck in there.

bokand:
@jakearchibald - latest commits should have addressed all your comments, WDYT?

annevk:
Thanks, that's fine. General approach seems good to me.

PR #239, opened 2023-11-21
[Spec] Overhaul 'restricting the text fragment' section

### LINKED ISSUE #179: Sec-Fetch-Site header inspection is wrong
You cannot inspect the `Sec-Fetch-Site` header as it's only appended just before the request hits the network. You cannot observe it from the code that deals with navigation.

bokand:
The spec was using the `sec-fetch-site` request header to determine the initiator properties of the navigation, to use in security restrictions. However, request headers are appended only just prior to performing the fetch, this part of the algorithm operates on a clone of the request without headers so this doesn't work.

This PR fixes the issues and rearranges the checks to occur in the `finalize a cross document navigation` and `navigate to a fragment` steps, also enabling text directives from a same-document navigation (the behavior in both Safari and Chrome, see #240).

Fixes #179

***
<a href="https://pr-preview.s3.amazonaws.com/bokand/ScrollToTextFragment/pull/239.html" title="Last updated on Nov 29, 2023, 4:11 PM UTC (81c42d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/scroll-to-text-fragment/239/2366cbc...bokand:81c42d9.html" title="Last updated on Nov 29, 2023, 4:11 PM UTC (81c42d9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1140,40 +1140,47 @@ boolean [=document/text directive user activation=] field:
   allowed in all cases.
 </div>
 
+Amend the definition of <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params">navigation params</a> to include a new field:
+
+>   <strong>Monkeypatching [[HTML]]:</strong>
+>
+>  <dl>
+>   <dt><dfn>user involvement</dfn></dt>
+>   <dd>A <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#user-navigation-involvement">user navigation involvement</a> value.</dd>
+>  </dl>
+>
+>
+> Initialize [=user involvement=] value everywhere a navigation params is created, in particular,
+> amend the definition of
+> <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#attempt-to-populate-the-history-entry's-document">attempt to populate the history entry's document</a>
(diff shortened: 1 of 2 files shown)


bokand:
@annevk does this look right? In particular, I noticed that the "create and initialize a Document object" steps [already check the source origin](https://html.spec.whatwg.org/multipage/document-lifecycle.html#initialise-the-document-object:~:text=and%20browsingContext%27s%20active%20document%27s%20origin) but in "navigate" we [snapshot](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate:~:text=Let%20initiatorOriginSnapshot%20be%20sourceDocument%27s%20origin.) the source origin so I'm guessing it can change and the existing case for about:blank is special?

bokand:
Ping - I've built a bit more on top of this so it no longer performs the checks in the 'create and initialize a document object' steps.

(Please see also questions in https://github.com/whatwg/html/issues/8282#issuecomment-1830061794, if you'd rather review once this is "ready" for PR into HTML I can do that too)

bokand:
@annevk in the interest of making progress, and since I have a number of other changes and fixes that are blocked on this, I'm going to merge this as-is. This will still be reviewed as part of the HTML PR. Happy to make changes post-merge if you find any.

PR #247, opened 2023-12-13
[Spec] Overhaul directive parsing

### LINKED ISSUE #220: Should FragmentDirective know about TextDirective?
I guess currently UnknownDirective is a superset of TextDirective but the current setup allows for them to get out of sync which doesn't seem desirable.

### LINKED ISSUE #221: Should the processing model care about validity?

### LINKED ISSUE #230: The spec is unclear on the relationship of percent-encoding and Unicode
The spec says:

> A text directive is a kind of [directive](https://wicg.github.io/scroll-to-text-fragment/#directives) representing a range of text to be indicated to the user. It is a [struct](https://infra.spec.whatwg.org/#struct) that consists of four strings: start, end, prefix, and suffix.

So _start_, _end_, _prefix_, and _suffix_ are defined as strings, though without a link to the Infra notion of "string".

The spec then says:

> Set retVal’s [prefix](https://wicg.github.io/scroll-to-text-fragment/#text-directive-prefix) to the [percent-decoding](https://url.spec.whatwg.org/#string-percent-decode) of the result of removing the last character from potential prefix.

[Percent-decode](https://url.spec.whatwg.org/#percent-decode) returns a byte sequence but the spec assigns the return value to _prefix_, which is a string. (Likewise for _start_, _end_, and _suffix_.)

The spec should say whether the byte sequence is converted to a string by applying [UTF-8 decode without BOM](https://encoding.spec.whatwg.org/#utf-8-decode-without-bom) or by applying [UTF-8 decode without BOM or fail](https://encoding.spec.whatwg.org/#utf-8-decode-without-bom-or-fail) (and what happens on failure if the latter).

bokand:
### Specify parsing imperatively

 * Add imperative steps to check validity and perform parsing
 * Make parsing happen earlier in the process so that we pass around parsed Text Directive objects rather than strings.
 * Make the steps more precise, referring to infra types and correctly decoding the strings.

Fixes #221
Fixes #230

### Fix and make grammar non-normative

* The grammar is now provided solely as a convenience and is non-normative.
* Make UnknownDirective not subsumed by TextDirective.

Fixes #220

***
<a href="https://pr-preview.s3.amazonaws.com/bokand/ScrollToTextFragment/pull/247.html" title="Last updated on Dec 13, 2023, 9:13 PM UTC (b406c89)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/scroll-to-text-fragment/247/5da963e...bokand:b406c89.html" title="Last updated on Dec 13, 2023, 9:13 PM UTC (b406c89)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -606,12 +606,10 @@ state=] to apply the directives associated with a session history entry to a [=/
 
 >   <strong>Monkeypatching [[DOM#interface-document]]:</strong>
 >
->   Each document has an associated <dfn for="Document">uninvoked directives</dfn> which is either
->   null or an ASCII string holding data used by the UA to process the resource. It is initially
->   null.
+>   Each document has an associated <dfn for="Document">pending text directives</dfn> which is either
+>   null or an <a spec=infra>list</a> of [=text directives=]. It is initially null.
 
-In the definition of <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#update-document-for-history-step-application">
-update document for history step application</a>:
+In the definition of <a spec="HTML">update document for history step application</a>:
 
 >   <strong>Monkeypatching [[HTML#updating-the-document]]:</strong>
 >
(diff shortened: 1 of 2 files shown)


bokand:
@annevk Similarly in this case, please let me know if you'd like me to change anything but otherwise I'll leave the review until the PR in HTML.

PR #254, opened 2024-02-02
Update README.md to fix typo

romain-gaillard:
The example in the text refers to the word `baz` but the URL contains `bas`. This fixes the typo in the URL so that it matches what's in the text.


bokand:
Thanks!

Issue #262, opened 2025-02-26
Quadratic behavior in find algorithm


jnjaeschke:
The spec introduces quadratic behavior when trying to find a text fragment that contains context terms (prefix, suffix).

The spec algorithm essentially does this:
- Search for the prefix string
- Create a new search range from the end of the prefix match to the end of the document
- Search for the `start` string in the search range
- If there is no match for `start` or the match for `start` is not at the beginning of the search range, discard the attempt

(same for start/end and suffix)

The consequences for this behavior can be experienced in this (albeit theoretical) use case:
- Go to this (arbitrary) [section](https://html.spec.whatwg.org/#revealing-the-document) of the HTML spec
- Create a text fragment for the first occurrence of the word "the" in that section
- Or, take this link: https://html.spec.whatwg.org/#revealing-the-document:~:text=the,-pagereveal%20event
- Open that link in Chrome (I used Chromium 133), Safari, Firefox Release (135) and Firefox Nightly (137)

Opening that link takes 15+ minutes on my dev machines due to quadratic behavior (Chrome, Safari and Firefox Release 135).

The Firefox Nightly implementation restricts the search range to the next block boundary instead of the end of the document (since the text directive context terms can't cross a block boundary). This reduces finding the text directive (again, on my dev machine) to about 400 milliseconds. 
Unless I'm missing something, the _correct_ solution could go even further and essentially do a `searchRange.startsWith(nextContextTerm)` and discard the attempt as soon as there's a mismatch. However, since I assume all implementations use the `window.find()` algorithm internally (and I further assume that those algorithms operate on ranges), restricting the search range to the nearest block boundary could be more pragmatic.

I'm aware that the spec technically is not _wrong_ here. But the fact that all implementations (minus Firefox 137) followed this behavior, which led to bad performance, might motivate to update the spec to eliminate the issue. I'm happy to provide a PR.

cc @zcorpan

bokand:
Thanks for pointing this out. Yeah, agree this is a perf-bug in the spec.

> The Firefox Nightly implementation restricts the search range to the next block boundary instead of the end of the document (since the text directive context terms can't cross a block boundary). 

Do you mean the next block boundary _after_ advancing to the next non-whitespace position? If so then yeah, I think that's reasonable. I don't recall why I chose to use a search from there, rather than just `startsWith` as you note...

Would be glad to review a PR if you'd like to put it up.

---

[...]

PR #263, opened 2025-03-19
Change the search algorithm as described in #262.

jnjaeschke:
This patch updates the `find a string in range` algorithm to use a new boolean parameter `matchMustBeAtBeginning`, which will early-return the algorithm if the `query` is not at the beginning of the `searchRange`.
This variant is used when searching for the `start` parameter when a `prefix` is present, and for the `suffix` parameter.

This also required some changes to the control flow of the `find a range from a text directive` algorithm.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1798,7 +1798,8 @@ following steps:
             1. Let |prefixMatch| be the the result of running the [=find a string
                 in range=] steps with |query| |parsedValues|'s
                 [=text directive/prefix=], |searchRange| |searchRange|,
-                |wordStartBounded| true and |wordEndBounded| false.
+                |wordStartBounded| true, |wordEndBounded| false and
+                |matchMustBeAtBeginning| false.
             1. If |prefixMatch| is null, return null.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1798,7 +1798,8 @@ following steps:
             1. Let |prefixMatch| be the the result of running the [=find a string
                 in range=] steps with |query| |parsedValues|'s
                 [=text directive/prefix=], |searchRange| |searchRange|,
-                |wordStartBounded| true and |wordEndBounded| false.
+                |wordStartBounded| true, |wordEndBounded| false and
+                |matchMustBeAtBeginning| true.
```
bokand, line 1802: It's been a while since I've looked at this so I might be missing something but shouldn't `matchMustBeAtBeginning` be false here? We're searching for the prefix here so setting this to true would only match a prefix that's at the start of the document.

jnjaeschke: Duh, of course. Sorry about that. Thanks for finding :)

bokand (approved):
LGTM - thanks!

